PR #7, opened 2017-02-25
Fix generating hyperrogue-music.txt

AMDmi3:
```sed -i``` is not portable. Use portable ```sed``` command instead to generated target file in one step

### CODE REVIEW

**Comment by nobrakal** [comment]
Thank you for your help :)

**Comment by AMDmi3** [comment]
Hmm, actually the target seems to be useless as it doesn't replace Maybe that's that's the cause 8.3j is broken, as while it installs music into ${datadir}/music, it searches for it in ${datadir}:

```
% hyperrogue
HyperRogue by Zeno Rogue <zeno@attnam.com>, version 8.3j
released under GNU General Public License version 2 and thus
comes with absolutely no warranty; see COPYING for details
Mix_LoadMUS: Couldn't open '/usr/local/share/hyperrogue/hr3-icyland.ogg'
% ls /usr/local/share/hyperrogue/**/hr3-icyland.ogg 
/usr/local/share/hyperrogue/music/hr3-icyland.ogg
```

PR #10, opened 2017-06-26
Respect user-set optimization flags

AMDmi3:
The build should respect optimization flags set by the user


nobrakal:
Thanks for catching that !

PR #11, opened 2017-06-26
Fix compilation error

AMDmi3:
You cannot perform ordered comparison on pointer and integer


AMDmi3:
clang output:

```
./sound.cpp:125:30: error: ordered comparison between pointer and zero ('char *' and 'int')
    while(fgets(buf, 800, f) > 0) {
          ~~~~~~~~~~~~~~~~~~ ^ ~
```

PR #15, opened 2017-10-17
Fix an accidental reversal of tree-chopping messages.

Quuxplusone:


nobrakal:
Hi,

Seems good to me :) I'm going to merge, @zenorogue please reverse if you think it was an error.

PR #16, opened 2017-10-17
Update "Makefile.mac" to work for OS X with Homebrew.

Quuxplusone:
And drive-by fixes for a couple of Xcode warnings in fa8078dc8b49d0705ff0f95be7bafd5d76d67ce6 — feel free to cherry-pick that one separately if you want it.

The original `Makefile.mac` was from circa 2011, so I claim that nobody should care about it anymore. :) But it would also be reasonable to keep the old `Makefile.mac` and add this one as `Makefile.osx` or something, if you want.


nobrakal:
Seems good to me. Thank you for your help ;)

Any idea @zenorogue ? Particularly in the first commit, I'm not a C dev...

zenorogue:
Thanks! I have also fixed the compiler warnings on my own for 10.1, but the new Makefile.mac will be definitely useful.

Issue #26, opened 2017-12-21
Pushing a thumper past a sandworm is more allowed than it should be


Quuxplusone:
When the player is adjacent to a `waThumperOn` and the thumper is adjacent to a `moWorm` (a sandworm ready to attack), it is possible for the player to push the thumper, even though the push would normally place the player adjacent to the attacking worm. In this situation, the thumper gets pushed (oops!) but then `checkmove()` prevents the player from moving, so the thumper ends up moving one space away from the player "for free" without any turn elapsing.

PR #27, opened 2017-12-21
Fixes issue #26

Quuxplusone:


zenorogue:
Older versions had a "You summon some Thumpers!" cheat, but I have decided to remove it (the map editor provides a more usable alternative, and better to keep the hotkey free). Unused functions removed, 'Z' improved, fixed the thumper bug.

PR #38, opened 2018-06-17
Minor fixes for Mac and MinGW.

Quuxplusone:
Eliminate a format-string warning by using less template magic:
https://stackoverflow.com/questions/12573968/how-to-use-gccs-printf-format-attribute-with-c11-variadic-templates

Fix one `%Ld` in `rogueviz.cpp` that should have been `%lld`.

Rename `savepng.c` into `savepng.cpp` so that we don't have to care about the system's C compiler; it builds fine using the same C++ options as HyperRogue itself.

Figure out how to get `-DCAP_PNG` and `-DCAP_ROGUEVIZ` working on Mac, and document them.
Assume that roughly the same things should also work on MinGW.

Consistency on `STDSIZE`. The one place where Clang still complains even with `-std=c++17 -DSTDSIZE -Wno-sign-compare` is here:

    ./fieldpattern.cpp:757:51: error: non-constant-expression cannot be narrowed from type 'unsigned long' to 'int' in initializer list
          [-Wc++11-narrowing]
          ex.primes.emplace_back(primeinfo{nextprime, size(fp.matrices) / S7, (bool) fp.wsquare});
                                                      ^~~~~~~~~~~~~~~~~~~~~~

So, we fix that up too, in this patch.

### DIFF
--- a/Makefile.mac
+++ b/Makefile.mac
@@ -1,4 +1,4 @@
-# This Makefile works for Mac OS X (Yosemite).
+# This Makefile works for Mac OS X (El Capitan).
 #
 # Run "brew install sdl" to install SDL in /usr/local.
 # Run "brew install sdl_gfx".
@@ -7,15 +7,31 @@
 # Run "make -f Makefile.mac" to build HyperRogue as ./hyper.
 
 CXXFLAGS += -std=c++11 -march=native -DMAC
-# CXXFLAGS += -DCAP_ROGUEVIZ
 CXXFLAGS += -W -Wall -Wextra -pedantic
-CXXFLAGS += -Wno-format-pedantic -Wno-unused-parameter -Wno-char-subscripts -Wno-missing-field-initializers -Wno-vla-extension
-CXXFLAGS += ${EXTRA_CXXFLAGS}
+CXXFLAGS += -Wno-format-pedantic -Wno-unused-parameter -Wno-missing-field-initializers -Wno-vla-extension
 CXXFLAGS += -I/usr/local/include
+CXXFLAGS += ${EXTRA_CXXFLAGS}
+
 LDFLAGS += -L/usr/local/lib
 
-hyper: hyper.o
-	$(CXX) $(CXXFLAGS) hyper.o $(LDFLAGS) -lSDL -lSDLMain -lSDL_gfx -lSDL_mixer -lSDL_ttf -framework AppKit -framework OpenGL -o hyper
+OBJS = hyper.o
+
+ifeq (a,b)
+# Enable PNG screenshots. Requires "brew install libpng".
+CXXFLAGS += -DCAP_PNG
+LDFLAGS += -lpng
+OBJS += savepng.o
+else
+CXXFLAGS += -DCAP_PNG=0
+endif
+
+ifeq (a,b)
+# Enable RogueViz.
+CXXFLAGS += -DCAP_ROGUEVIZ
+endif
+
+hyper: $(OBJS)
+	$(CXX) $(CXXFLAGS) $(OBJS) $(LDFLAGS) -lSDL -lSDLMain -lSDL_gfx -lSDL_mixer -lSDL_ttf -framework AppKit -framework OpenGL -o hyper
 
 hyper.o: *.cpp language-data.cpp
 	$(CXX) $(CXXFLAGS) -O2 -c hyper.cpp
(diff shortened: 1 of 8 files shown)


zenorogue:
Thanks! I have replaced `using namespace std` with an explicit list of `using std::...`, which does not include `std::size` -- I believe this should fix this issue, and potentially similar issues in the future. Thus STDSIZE is removed.

Quuxplusone:
> I have replaced `using namespace std` with an explicit list of `using std::...`, which does not include `std::size` -- I believe this should fix this issue, and potentially similar issues in the future. Thus STDSIZE is removed.

Thanks, that is a huge improvement! (C++ pedants [hate](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice) `using namespace std`. 🙂)  However, the problem here is actually *not* related to `using`; the problem is ADL. When you call *unqualified* `size` on a `std::vector`, name lookup is going to look in `namespace std` because that is one of the "associated namespaces" of `std::vector`. This is going to happen regardless of any `using`-directives in the program — `using`-directives can only *add* namespaces into scope, not *remove* them.

So in master right now, `-std=c++17` (still) gives errors:

```
langen.cpp:126:18: error: call to 'size' is ambiguous
  for(int i=0; i<size(s); i++)
                 ^~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iterator:1584:16: note: candidate
      function [with _Cont = std::__1::basic_string<char>]
constexpr auto size(const _Cont& __c) -> decltype(__c.size()) { return __c.size(); }
               ^
langen.cpp:22:23: note: candidate function [with T = std::__1::basic_string<char>]
template<class T> int size(const T& x) { return x.size(); }
                      ^
```

And then there's another subtle point making this hard to fix non-invasively: your global `size` returns signed `int`, but `std::size` returns unsigned `size_t` (because that's `std::vector::size_type`). So if you just ifdef-out your global `size` template, everything compiles, but you get a whole slew of warnings from `-Wsign-compare`:

```
langen.cpp:126:17: warning: comparison of integers of different signs: 'int' and 'decltype(__c.size())' (aka 'unsigned long') [-Wsign-compare]
  for(int i=0; i<size(s); i++)
               ~^~~~~~~~
```

There are three idiomatic ways to fix *that*:
- Add `-Wno-sign-compare` to the Makefiles. This silences the warning, but it means you'll miss any *future* instances of this warning that might actually be telling you about bugs. Right now the codebase is actually `-Wsign-compare`-clean, which is awesome.
- Change `i` from `int` to `size_t`. This changes the meaning of expressions like `-i` and `i-1`, though, so you should be very scared of doing this by search-and-replace.
- Change this for-loop to a range-based for-loop: `for (char ch : s)`. This would actually be awesome, and is relatively safer, but it absolutely *cannot* be done by mechanical search-and-replace; it *must* be done by hand on a case-by-case basis.

My [proposed solution](https://github.com/Quuxplusone/hyperrogue/commit/cpp17) does mechanical search-and-replace with carefully crafted regexes that I believe (and you could double-check) *never* change the meaning of the expression. The downside is that the change inserts a lot of `(int)` casts. About half of them are redundant. But you could remove the redundant casts gradually, on a case-by-case basis — just any time you touch a piece of code with redundant casts, remove them from that piece of code (and maybe add a ranged for-loop, too!).

Finally, there are two ways to solve the problem by avoiding the ADL collision in the first place:
- Search-and-replace `size(x)` into `mysize(x)` (or `Size`, or `hrsize`, or `sz`). ADL will still happen, but now there is no `std::mysize`/`std::Size`/`std::hrsize`/`std::sz` to collide with, so you'd be okay.
- Search-and-replace `size(x)` into `::size(x)`, explicitly qualifying that you mean the `size` in the global namespace. This disables ADL lookup entirely, so you'd be okay.

The only reason I don't like these solutions is that I don't think you should have the `size` function to begin with — I'd personally rather write `x.size()` and then cast it to a different type (`int`) iff I wanted it to be that different type. So that's the patch I'm proposing to supply.

zenorogue:
Thanks! Indeed, my solution does not work correctly. I have decided to renamed the `size` function to `isize` (i.e., size casted to int -- this name suggests how this is different from std::size). I like `hr::size`, but I think it is dangerous, because I would probably be likely to forget `hr::`, causing potential problems in C++17. In my opinion `(int)x.size()` looks rather ugly (especially when casting is not necessary) and I prefer to use a function.

Speaking of loops using `size` instead of range-based loops -- this is usually for one of two reasons:
(1) They were written before HyperRogue embraced C++11 (and changing them would be indeed awesome, though probably not worth the chance that some bugs might be introduced during replacement). These are usually replaced when working on the particular function :)
(2) They need to access the index, and I do not see a way that I would consider more elegant. Does the [zip solution mentioned in your blog](https://quuxplusone.github.io/blog/2018/06/17/std-size/) work correctly? You seem to be accessing tuple elements via ld[0] and ld[1]. Structured binding in C++17 would be perfect, but better to remain compatible with C++11 for now.

PR #41, opened 2018-06-25
Fix build by specifying correct savepng.cpp file name

AMDmi3:
Please consider adding a CI (such as [Travis](https://travis-ci.org/)) support, as these simple but fatal errors just keep crawling in.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,7 +3,7 @@ bin_PROGRAMS = hyperrogue
 hyperroguedir = $(datadir)/hyperrogue
 dist_hyperrogue_DATA = hyperrogue-music.txt DejaVuSans-Bold.ttf
 dist_doc_DATA = README.md
-hyperrogue_SOURCES = hyper.cpp savepng.c
+hyperrogue_SOURCES = hyper.cpp savepng.cpp
 
 # Some definitions used in graph.cpp
 hyperrogue_CPPFLAGS = -DFONTDESTDIR=\"$(pkgdatadir)/DejaVuSans-Bold.ttf\" -DMUSICDESTDIR=\"$(pkgdatadir)/hyperrogue-music.txt\" -DSOUNDDESTDIR=\"$(pkgdatadir)/sounds/\"


Quuxplusone:
This one was my fault (db3e5d10090eae170001bfe37dbcd3b2e722dc99); I fixed the Makefiles, which I'm able to test/read, but I have no idea how to test whatever build process takes `Makefile.am`as its input.
The proposed patch LGTM.

Quuxplusone:
@AMDmi3: Could you post a complete, cut-and-pasteable script that builds `./hyper` for you on some flavor of Linux and/or OSX?  I have been trying and failing for the last hour or two to get it working on Ubuntu via Travis-CI:

https://github.com/Quuxplusone/hyperrogue/commit/d1fe18c6
https://travis-ci.org/Quuxplusone/hyperrogue/jobs/396694260 (might not be publicly visible, I'm not sure)

The output once it finally starts to run `make` is:

```
clang++ -DPACKAGE_NAME=\"hyperrogue\" -DPACKAGE_TARNAME=\"hyperrogue\"
 -DPACKAGE_VERSION=\"10.4b\" -DPACKAGE_STRING=\"hyperrogue\ 10.4b\"
 -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"hyperrogue\"
 -DVERSION=\"10.4b\" -DHAVE_LIBGL=1 -DHAVE_LIBSDL=1 -DHAVE_LIBSDL_MIXER=1
 -DHAVE_LIBSDL_GFX=1 -DHAVE_LIBSDL_TTF=1 -DHAVE_LIBPNG=1 -DHAVE_LIBGLEW=1
 -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1
 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1
 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_STDLIB_H=1 -DHAVE_UNISTD_H=1
 -DHAVE_SDL_SDL_H=1 -DHAVE_SDL_SDL_MIXER_H=1 -DHAVE_SDL_SDL_TTF_H=1
 -DHAVE_MEMSET=1 -DHAVE_STRSTR=1 -I.    -O0  -MT langen-langen.o
 -MD -MP -MF .deps/langen-langen.Tpo -c -o langen-langen.o `test -f 'langen.cpp' || echo './'`langen.cpp
langen.cpp:28:46: error: no member named 'move' in namespace 'std'
    if(m.count(s)) add(s + " [repeat]", std::move(val));
                                        ~~~~~^
```

That is, `-std=c++11` doesn't appear anywhere in `AM_CXXFLAGS` nor `langen_CXXFLAGS` at the moment, and therefore `langen.cpp` doesn't compile.

----
I propose that I could write a `Makefile` along the same lines as `Makefile.mac` and `Makefile.mgw` in about half an hour, but I'm not sure — might we be getting some hypothetical future benefit from using GNU autotools here?

One possible benefit of autotools would be if you could `./configure --disable-glew` to keep `./configure` from crashing out on platforms that don't have GLEW installed; but I don't see that in the current codebase.

Or, if the codebase actually used autotools' `HAVE_FOO` macros, then it would make sense that we were bothering to generate all of them. But the codebase currently uses `CAP_FOO` macros instead.

Or, if the codebase used autotools' slew of `PACKAGE_FOO` macros; but it doesn't.

AMDmi3:
It turns out that landgen needs to set C++ standard as well. See #43 - this is working Travis build.

PR #47, opened 2018-06-29
Eliminate VLAs; eliminate std::random_shuffle()

Quuxplusone:
These two commits are independent. They make the code more MSVC-friendly by eliminating things that don't exist in C++ (VLAs) or don't exist in C++17 (`random_shuffle`).

### DIFF
--- a/Makefile.mac
+++ b/Makefile.mac
@@ -8,7 +8,7 @@
 
 CXXFLAGS += -std=c++11 -march=native -DMAC
 CXXFLAGS += -W -Wall -Wextra -pedantic
-CXXFLAGS += -Wno-format-pedantic -Wno-unused-parameter -Wno-missing-field-initializers -Wno-vla-extension
+CXXFLAGS += -Wno-format-pedantic -Wno-unused-parameter -Wno-missing-field-initializers
 CXXFLAGS += -I/usr/local/include
 CXXFLAGS += ${EXTRA_CXXFLAGS}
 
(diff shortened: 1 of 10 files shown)


zenorogue:
Thanks! I do not understand the comment about the generalization of 'hrandom_shuffle' -- this does not seem to be done?

Regarding 'randshuffle' vs 'hrandshuffle' -- I have added a comment about when 'hrand' should be used, since this was not explained anywhere:

```// the main random number generator for the game
// all the random calls related to the game mechanics (land generation, AI...) should use hrngen
// random calls not related to the game mechanics (graphical effects) should not use hrngen
// this ensures that the game should unfold exactly the same if given the same seed and the same input```

This is useful for debugging, and for some potential game recording feature. In RogueViz it is also useful to generate the same result if the same seed is given.

So using `random_shuffle` instead of `hrandom_shuffle` in `game.cpp` was a bug, but using  `random_shuffle` in `rogueviz-kohonen.cpp` was correct (it was done in a function generating a right-click description of a cell, so right-clicking while a SOM was still generated would change the state of the RNG).

Quuxplusone:
> Thanks! I do not understand the comment about the generalization of 'hrandom_shuffle' -- this does not seem to be done?

Oops. I started with a change that changed `hrandom_shuffle` and added a call to `hrandom_shuffle(v.begin(), v.size())`. After writing that commit message, I decided to grep and compare uses of `vec.begin()` to `&vec[0]`, and discovered that you actually *were* writing `&vec[0]` consistently; so I just changed my one new use of `v.begin()` to `&v[0]` and undid my change to `hrandom_shuffle`. And then I forgot to change my commit message. Sorry about that.

PR #48, opened 2018-06-30
[ISSUE IN THE FORM OF A PATCH] MSVC warns about bool/int conversion

Quuxplusone:
This patch does, technically, fix the following MSVC diagnostics.
But I believe that it should *not* be committed exactly as-is.
I suspect that some of these diagnostics — especially the ones in `mapeditor.cpp` — might indicate possible bugs or at least tech debt. Committing this patch as-is would just make the bug/debt harder to find in the future. So I'd be happier if you left this PR open until each instance is fixed in some "good" way, and then closed this PR without merging it.

There is also a trivial workaround for MSVC: I can just pass `/wd4805`, the same way I pass `-Wno-unused-result` on GCC.

    heptagon.cpp(201): warning C4805: '^': unsafe mix of type 'const bool' and type 'int' in operation
    pattern2.cpp(627): warning C4805: '^=': unsafe mix of type 'int' and type 'bool' in operation
    pattern2.cpp(669): warning C4805: '^=': unsafe mix of type 'int' and type 'bool' in operation
    mapeditor.cpp(1325): warning C4805: '!=': unsafe mix of type 'int' and type 'bool' in operation
    mapeditor.cpp(1770): warning C4805: '==': unsafe mix of type 'bool' and type 'int' in operation
    hypgraph.cpp(583): warning C4805: '^': unsafe mix of type 'const bool' and type 'int' in operation
    hyper.h(2025): warning C4805: '!=': unsafe mix of type 'bool' and type 'int' in operation
    graph.cpp(3464): note: see reference to function template instantiation 'V hr::callhandlers<bool(hr::cell *,const hr::transmatrix &),int,hr::cell*,hr::transmatrix>(V,hr::hookset<bool (hr::cell *,const hr::transmatrix &)> *,hr::cell *&,hr::transmatrix &)' being compiled with [V=int]

### DIFF
--- a/graph.cpp
+++ b/graph.cpp
@@ -3461,7 +3461,7 @@ void drawcell(cell *c, transmatrix V, int spinv, bool mirrored) {
 
   // if(behindsphere(V)) return;
   
-  if(callhandlers(0, hooks_drawcell, c, V)) return;
+  if(callhandlers(false, hooks_drawcell, c, V)) return;
   
   ld dist0 = hdist0(tC0(V)) - 1e-6;
   if(dist0 < geom3::highdetail) detaillevel = 2;
--- a/heptagon.cpp
+++ b/heptagon.cpp
@@ -198,7 +198,7 @@ heptspin operator + (const heptspin& hs, wstep_t) {
   createStep(hs.h, hs.spin);
   heptspin res;
   res.h = hs.h->move[hs.spin];
-  res.mirrored = hs.mirrored ^ hs.h->mirror(hs.spin);
+  res.mirrored = boolean_xor(hs.mirrored, hs.h->mirror(hs.spin));
   res.spin = hs.h->spin(hs.spin);
   return res;
   }
--- a/hyper.h
+++ b/hyper.h
@@ -421,6 +421,8 @@ int hrand(int i);
 // size casted to int, to prevent warnings and actual errors caused by the unsignedness of x.size()
 template<class T> int isize(const T& x) {return x.size(); }
 
+bool boolean_xor(bool a, bool b) { return a ^ b; }
+
 // initialize the achievement system.
 void achievement_init();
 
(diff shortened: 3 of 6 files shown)


zenorogue:
In some cases type int was used when bool should have been used (the return type of mirror (and tmirror), the argument of callhandlers).

In the other cases, it assumes that bools are converted to 0 or 1, which is specified by the standard AFAIK, so I do not think that the warning is useful. I have added an explicit conversions anyway.

Are all the warnings fixed now?

BTW hyper.h is a header file, so all functions defined there should be template or inline. (For experiments/animations/trailers/etc, I write a cpp file which defines some hooks, includes *.h and links with hyper.o -- this does not work when a function is "defined twice")

Quuxplusone:
> I do not think that the warning is useful

Yeah, personally I approve of your `x ^ y` uses and wish that didn't trigger the same warning in MSVC. But the blanket warning did find some true-positives, too.

> hyper.h is a header file, so all functions defined there should be template or inline.

Oops! (And the hooks thing makes sense, because I've noticed that there are several hooksets that are completely unused in master but must be used behind the scenes.)

> Are all the warnings fixed now?

Possibly, but the Windows build is failing now on a new use of `max()` in rogueviz.cpp, which I'll have to google how to shut up, which probably won't happen until late tonight. More news on that later.

Meanwhile , there's a new error on OSX Clang:
```
error: comparison of constant 103 with expression of type 'hr::eGeometry' is always true
      [-Werror,-Wtautological-constant-out-of-range-compare]
    if(targetgeometry != rg::geometry)
       ~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~
```
I can't quite figure out what's going on there. I think maybe you meant to compare to regular `geometry` instead of `rg::geometry`? (Incidentally, for the `rg::` constants you could be using an `enum { geometry = 'g', bitrunc = '7',  gp = 'w',` etc., instead of static chars with their own symbols/addresses.)

Travis-CI actually [would have caught this now, kinda](https://travis-ci.org/zenorogue/hyperrogue/jobs/401228605), but the autotools stuff doesn't pass `-Werror` so the warning was swallowed. I wonder if @AMDmi3 could submit a patch to add `-Werror` on autotools builds?

Quuxplusone:
> Are all the warnings fixed now?

Never mind my pessimism; I can now confirm that yes all the bool-to-int related warnings are fixed in master. :)

zenorogue:
Yes, this should have been `geometry`. Thanks!

PR #50, opened 2018-07-07
Enable -Werror in Travis

AMDmi3:
As requested at https://github.com/zenorogue/hyperrogue/pull/48#issuecomment-403230315


Quuxplusone:
You need something like `-Wno-unused-result`, too.
I'd say, check the Makefile and do whatever it does.

AMDmi3:
Done. There's still warning that should be fixed.

PR #53, opened 2018-07-11
Merge Makefiles. Add emscripten to Makefile. Test all platforms on TravisCI.

Quuxplusone:
There is now a single `Makefile.simple` to replace `Makefile.mac` and `Makefile.mgw`. It works on all three Unix-ish platforms (Mac, MinGW, and Linux), and also can be used to build the emscripten build via any of the following command lines:

    make -f Makefile.simple emscripten
    make -f Makefile.simple hyper.html
    make -f Makefile.simple hyper.js
    make -f Makefile.simple hyper.wasm

`.travis.yml` now tests `Makefile.simple` on all platforms Travis knows about. This includes testing the `-DCAP_ROGUEVIZ=1` and `-DCAP_PNG=0` builds, which is not possible with autotools.

Finally, with `-DCAP_ROGUEVIZ` we find a bug in rogueviz.cpp: `.2` is not a valid step size for an `int` slider. So I patched that.


zenorogue:
Thanks! Looks great. Step size .2 was meaningful since this slider uses logarithmic scale, but of course it did not work as it was written -- fixed.

PR #55, opened 2018-07-14
Fix -Wempty-body warnings on GCC.

Quuxplusone:
Turns out there are only three of these warnings.


zenorogue:
IIRC this has been written like this to avoid the "unused result" warning... I think it would be better to handle  the possible scanf errors correctly (or at least pretend that they are handled in a better way).

PR #56, opened 2018-07-18
Integrate with Appveyor CI

Quuxplusone:
I'm not sure of the cost/benefit here, but there's no harm in posting a pull request. :)

Benefit: Appveyor will test the MinGW and MSVC builds!
Cost: 120 lines of Appveyor-specific .yml and .bat files. Also, you'll have to give GitHub permissions to Appveyor — and it asks for more permissions than Travis, in suspiciously broken English. :confused:

I'd also want to hear from @Danfun64 that I'm not messing up his MinGW build with this change to Makefile.simple.

`-DMSVC_COMPILER_LIMITS` is only because MSVC is stupid (and non-conforming). I haven't thought of a better *easy* way to fix that fatal error. I think better but *complicated* ways would be to rewrite that function as some kind of lookup into a big table of function pointers (expressed as lambdas), or break it down into subfunctions at the points currently indicated by comments (`// cheats`, `// mode changes:`, `// informational:`, etc).


zenorogue:
I have broken down the big branching function by moving some of the commandline options to the modules they belong to. Hopefully the parts are OK.

Danfun64:
Somehow I completely missed this! I'll have to check this in a few hours...

Quuxplusone:
Update: MSVC seems to be able to compile hyper.cpp without hitting that internal limit, which is cool. But now MSVC's linker is unhappy for a different reason. I'll look into this.
````
C:\projects\hyperrogue\hyper.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
````
In the meantime, I think it would be safe to cherry-pick efc9e81 if you wanted.

Quuxplusone:
Confirmed — `-bigobj` fixes the linker error. AFAIC, it's safe to merge this PR if you want it (except that I suppose it's still pending confirmation from @Danfun64 about the MinGW build).

Danfun64:
MSYS2/MinGW-W64 is confirmed to work with the new build.

Quuxplusone:
Rebased.

Quuxplusone:
TravisCI supports Windows now! But I haven't figured out how to build HyperRogue on it yet.
https://blog.travis-ci.com/2018-10-11-windows-early-release
https://docs.travis-ci.com/user/reference/windows

PR #67, opened 2019-01-06
Replace std::function with hyper_function for a 37% speedup in compile time

Quuxplusone:
commit f6f96e1129df134a9921015dc6e59d45ce5a7e11

    Fix some minor std::function-related bugs.
    
    Lambdas to be stored in `function<void()>` should not return `bool`.
    
    Two uses of `std::function` could be just `function`, like everywhere
    else in the codebase.

commit 4b3b2fd15c7cfc1be3202e290eff3b654d829fae

    Replace std::function with hyper_function for a 37% speedup in compile time.
    
    Before:
    
        time c++ -O2 -DMAC -I/usr/local/include -std=c++11 -march=native
                 -W -Wall -Wextra -Werror -pedantic -Wno-format-pedantic
                 -Wno-missing-field-initializers -Wno-unused-parameter
                 -DCAP_GLEW=0 -DCAP_PNG=0    -c hyper.cpp -o hyper.o
    
        real 2m22.508s
        user 2m20.625s
        sys  0m1.648s
    
    After:
    
        time c++ -O2 -DMAC -I/usr/local/include -std=c++11 -march=native
                 -W -Wall -Wextra -Werror -pedantic -Wno-format-pedantic
                 -Wno-missing-field-initializers -Wno-unused-parameter
                 -DCAP_GLEW=0 -DCAP_PNG=0    -c hyper.cpp -o hyper.o
    
        real 1m30.515s
        user 1m29.793s
        sys  0m0.689s
    
    Comparing object file size:
    
        -rw-r--r--  1 ajo  staff  8215036 Jan  5 20:46 old-hyper.o
        -rw-r--r--  1 ajo  staff  7547232 Jan  5 20:47 new-hyper.o
    
    Comparing number of symbols:
    
        nm old-hyper.o | wc -l  => 12590
        nm new-hyper.o | wc -l  =>  9742
    
    No appreciable difference in link time; the linker takes less than
    half a second in either case.

### DIFF
--- a/dialogs.cpp
+++ b/dialogs.cpp
@@ -473,7 +473,7 @@ namespace dialog {
   
   color_t *colorPointer;
   
-  bool handleKeyColor(int sym, int uni) {
+  void handleKeyColor(int sym, int uni) {
     unsigned& color = *colorPointer;
     int shift = colorAlpha ? 0 : 8;
 
@@ -514,7 +514,6 @@ namespace dialog {
     else if(doexiton(sym, uni)) {
       popScreen();
       }
-    return false;
     }
   
   bool colorAlpha;
--- a/graph.cpp
+++ b/graph.cpp
@@ -1840,9 +1840,9 @@ int taildist(cell *c) {
   }
 
 int last_wormsegment = -1;
-vector<vector< std::function<void()> > > wormsegments;
+vector<vector< function<void()> > > wormsegments;
 
-void add_segment(int d, const std::function<void()>& s) {
+void add_segment(int d, const function<void()>& s) {
   if(isize(wormsegments) <= d) wormsegments.resize(d+1);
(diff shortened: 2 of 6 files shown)


Quuxplusone:
Interestingly, I did a global sed to replace `= []` with `= +[]` and the compile time went back *up* from 1m30s to 1m50s (even as the number of distinct symbols in the `nm` output dropped from 9742 to 9132). I don't know how to explain that increase in compile time.

I've verified that this change would be acceptable to MSVC according to the AppVeyor build; although the AppVeyor build is currently broken for some reason related to SDL include paths.

zenorogue:
Wow, that is quite a big improvement. Although on my computer it is much less substantial (~2%). It does not compile on some older compilers I am using, so I have added USE_STDFUNCTION for now. Thanks!

PR #70, opened 2019-01-16
Fix colors for big endian architectures

julianaito:
Hi, 

It has been noticed by @ibara that colors are off when playing HyperRogue on a big endian machine. Blue was red etc. (see https://marc.info/?l=openbsd-ports&m=154739730704091&w=2). 

Here is a PR that fixes the issues in GL and textured mode.

### DIFF
--- a/polygons.cpp
+++ b/polygons.cpp
@@ -762,7 +762,12 @@ void fixMercator(bool tinf) {
   }
   
 unsigned char& part(color_t& col, int i) {
-  unsigned char* c = (unsigned char*) &col; return c[i];
+  unsigned char* c = (unsigned char*) &col;
+#if SDL_BYTEORDER == SDL_BIG_ENDIAN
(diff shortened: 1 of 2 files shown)


Quuxplusone:
IMHO it would be better to replace `unsigned char& part(color_t, int)` with separate "getter and setter" methods

    unsigned char part(color_t col, int i) {
        return col >> (8 * i);
    }

    void setpart(color_t& col, int i, unsigned char val) {
        mask = 0xFF << (8 * i);
        col = (col & ~mask) | (val << (8*i));
    }

thus entirely eliminating the dependency on endianness. But that would be a bigger patch, and I don't know whether it would be accepted any more than this one.

julianaito:
As long as the issue is fixed on your side, i'm fine with it, so we can
drop downstream patches :)

zenorogue:
Thanks! Merged (with some modifications).

PR #71, opened 2019-01-25
Avoid casting to `unsigned char*` in color calculations.

Quuxplusone:
Fixes the same thing as #70, but in a different way.

### DIFF
--- a/basegraph.cpp
+++ b/basegraph.cpp
@@ -851,7 +851,7 @@ void addMessage(string s, char spamtype) {
 
 color_t colormix(color_t a, color_t b, color_t c) {
   for(int p=0; p<3; p++)
-    setpart(a, p, part(a,p) + (part(b,p) - part(a,p)) * part(c,p) / 255);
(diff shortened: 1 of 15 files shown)


Quuxplusone:
@zenorogue ping? Any chance of getting something like this into master?

zenorogue:
I like the lvalue approach much more than the getter/setter/switcher approach... so unless there is some serious problem with that, I would like to keep it. E.g., change it so that e.g. part(color, 0) returns the first byte on LE and the last byte on BE; or use a proxy class.

Quuxplusone:
I would be happy to submit a patch that makes `part(c, i)` return a proxy class with an `operator=` so that the callsites could keep on using `part(c, i) = value`, but that seemed like a higher level of "C++ cleverness" than HyperRogue's normal level. I guess I should submit that PR and then we can compare that PR to this one (and to #70).

zenorogue:
Oh, I did not notice #70 did use my first suggested method. Merged that, also included some of your suggestions. Thanks!

PR #74, opened 2019-02-08
Rearrange some macros. No functional change intended.

Quuxplusone:
Consistently use `#if`, not `#ifdef`, for `ISSTEAM`. Move its initialization into the same place as `ISMAC`, `ISFAKEMOBILE`, etc.

Rearrange the `USE_STDFUNCTION` macros for straightforwardness.

Eliminate an obsolete check for `#ifdef INF` which is always true at this point.

### DIFF
--- a/hyper_function.h
+++ b/hyper_function.h
@@ -1,10 +1,24 @@
 #pragma once
 
+#if USE_STDFUNCTION
+
+#include <functional>
+namespace hr {
+    using std::function;
+} // namespace hr
+
+#else
+
 #include <type_traits>
 #include <utility>
 
+namespace hr {
+
 template<class Sig>
 class hyper_function;
+
+template<class Sig>
+using function = hyper_function<Sig>;
 
 template<class R, class... Args>
 struct hyper_function_state_base {
@@ -59,3 +73,7 @@ public:
         return ptr_ != nullptr;
     }
 };
+
+} // namespace hr
+
+#endif
--- a/commandline.cpp
+++ b/commandline.cpp
@@ -39,7 +39,7 @@ eMonster readMonster(const string& ss) {
 void initializeCLI() {
   printf("HyperRogue by Zeno Rogue <zeno@attnam.com>, version " VER "\n");
 
-#ifndef NOLICENSE
+#if !NOLICENSE
   printf("released under GNU General Public License version 2 and thus\n");
   printf("comes with absolutely no warranty; see COPYING for details\n");
 #endif
--- a/help.cpp
+++ b/help.cpp
@@ -137,7 +137,7 @@ void buildCredits() {
   help += XLAT("game design, programming, texts and graphics by Zeno Rogue <zeno@attnam.com>\n\n");
   if(lang() != 0)
     help += XLAT("add credits for your translation here");
-#ifndef NOLICENSE
+#if !NOLICENSE
   help += XLAT(
     "released under GNU General Public License version 2 and thus "
     "comes with absolutely no warranty; see COPYING for details\n\n"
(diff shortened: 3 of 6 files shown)


hyper_function.h
```diff
@@ -1,11 +1,25 @@
 #pragma once
 
+#if USE_STDFUNCTION
```
Quuxplusone: I'd be interested to see what error message you're getting from the compiler when `ISSTEAM && !USE_STDFUNCTION`. Maybe it's something easily fixable. Even though I see a huge speedup from `hyper_function`, I don't think it makes sense to keep it if it requires maintaining two different codepaths in perpetuity.
It would also be really nice to get the `ISSTEAM` build running on TravisCI (e.g. the way Emscripten is, using a Docker image), but I guess that might not be possible?

Orthogonally, you're right, the difference in compile time between `std::function` and `hyper_function` seems to be Clang-specific. E.g. I see a big difference between [202](https://travis-ci.org/zenorogue/hyperrogue/jobs/483899113) and [151](https://travis-ci.org/zenorogue/hyperrogue/jobs/490646163) seconds (Clang/Linux) but no difference (or worse than no difference) between [157](https://travis-ci.org/zenorogue/hyperrogue/jobs/483899114) and [173](https://travis-ci.org/zenorogue/hyperrogue/jobs/490646164) seconds (GCC/Linux).

eryxcc: Yes, probably they are easy to fix. I get the following errors:
```
hyper_function.h:13:44: error: 'virtual hyper_function_state_base<R, Args>::~hyper_function_state_base()' declared virtual cannot be defaulted in the class body
hyper_function.h:18:11: error: expected nested-name-specifier before 'Self'
hyper_function.h:18:11: error: using-declaration for non-member at class scope
hyper_function.h:18:16: error: expected ';' before '=' token
hyper_function.h:18:16: error: expected unqualified-id before '=' token
hyper_function.h:21:26: error: expected ';' at end of member declaration
hyper_function.h:21:32: error: 'override' does not name a type
hyper_function.h:24:52: error: expected ';' at end of member declaration
hyper_function.h:24:58: error: 'override' does not name a type
hyper_function.h:32:51: sorry, unimplemented: non-static data member initializers
hyper.h:22:21: error: expected unqualified-id before 'using'
```

zenorogue: Yes, probably they are easy to fix. I get the following errors:
```
hyper_function.h:13:44: error: 'virtual hyper_function_state_base<R, Args>::~hyper_function_state_base()' declared virtual cannot be defaulted in the class body
hyper_function.h:18:11: error: expected nested-name-specifier before 'Self'
hyper_function.h:18:11: error: using-declaration for non-member at class scope
hyper_function.h:18:16: error: expected ';' before '=' token
hyper_function.h:18:16: error: expected unqualified-id before '=' token
hyper_function.h:21:26: error: expected ';' at end of member declaration
hyper_function.h:21:32: error: 'override' does not name a type
hyper_function.h:24:52: error: expected ';' at end of member declaration
hyper_function.h:24:58: error: 'override' does not name a type
hyper_function.h:32:51: sorry, unimplemented: non-static data member initializers
hyper.h:22:21: error: expected unqualified-id before 'using'
```

Quuxplusone: Looks like GCC 4.6.
So the Steam build must not use either `Makefile.simple` or `Makefile.am` right now, because those makefiles both expect `-std=c++11` to be supported, whereas GCC 4.6 requires `-std=c++0x`, is that right?
I've got a branch at `Quuxplusone/steam` [(TravisCI)](https://travis-ci.org/Quuxplusone/hyperrogue/builds/491355022) which addresses the issues with GCC 4.6.4 and starts testing GCC 4.6.4 on Travis. Is that useful to you? and is there anything I could add to that build (e.g. `-DISSTEAM=1`) that would make it even more "realistic"?

PR #75, opened 2019-04-03
git rm some ancient files. NFC.

Quuxplusone:
`zeno.css` seems to be incomplete: it ends in the middle of a `{` block.

`hyperrogue.html` is the only user of `zeno.css`.

`setversion-sh` looks potentially useful, but has bit-rotted: it thinks `#define VER` should be in `hyper.cpp` when really it's been in `hyper.h` for a long time. So it's clearly not being used in practice.

I suggest removing all these files from top-of-tree. If people want them for historical interest, they can still be found in the git history.


zenorogue:
Thanks! Yeah, I think removing `zeno.css` and `hyperrogue.html` is a good idea -- `README.md` should fill their purpose (although `README.md` should be updated... it still gives $175 as the best score, so it has not been checked for a long time).

The purpose of `setversion-sh` is to make the executable version information in Windows correct; I use it for Windows binary releases, but some work would be necessary to integrate it in the build process correctly, and it is not that important. So I agree with deleting it for now.

PR #85, opened 2019-11-28
Slightly improved russian translation

shnourok:


eryxcc:
Thanks! By the way, I have lost contact with the original Russian translator, so the Russian translation is stuck at 10.3n. It would be helpful to translate the new additions.

zenorogue:
Thanks! By the way, I have lost contact with the original Russian translator, so the Russian translation is stuck at 10.3n. It would be helpful to translate the new additions.

PR #88, opened 2020-02-14
Fix several minor issues causing TravisCI to be red

Quuxplusone:
However, even with these fixed, TravisCI is _still_ red for the GCC 4.6 `-std=c++0x` build and for the Emscripten build.

I would like to get those banners back to green. Please let me know if you'd like to see patches in that direction. E.g.:

- Is `gcc-4.6` too old at this point? Shall we just retire it and move up to GCC 4.9.4-or-later?

- Shall we move up to C++14 at the same time? (I'm surprised that no C++14isms have crept in, even with the build red since e925f6e93ca358896bd4c2d7a73979d5fadb6d45.)

- Shall I invest any effort in sorting out why GCC 4.6 doesn't like `discoveries[k]`, and why each `discovery` needs a mutex in the first place, and whether the code still works if you unset `CAP_THREAD`? (HyperRogue does still compile on OSX if you set `-DCAP_THREAD=0`, for whatever that's worth.)

- Shall I invest any effort in figuring out how to use zlib on Emscripten? (So far, I don't even understand why it's needed.)

### DIFF
--- a/Makefile.simple
+++ b/Makefile.simple
@@ -63,7 +63,7 @@ ifeq (${OS},mingw)
   LDFLAGS_GL := -lopengl32
   LDFLAGS_GLEW := -lglew32
   LDFLAGS_PNG := -lpng
-  LDFLAGS_SDL := -lSDL -lSDL_gfx -lSDL_mixer -lSDL_ttf -lz -lphtread
+  LDFLAGS_SDL := -lSDL -lSDL_gfx -lSDL_mixer -lSDL_ttf -lpthread -lz
   OBJ_EXTENSION := .o
   hyper_RES := hyper.res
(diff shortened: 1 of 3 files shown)


zenorogue:
Thanks for your help!

I have upgraded my compiler for Steam Linux to gcc 5.4.1, and MinGW to 8.2.0. So gcc 5.4.1 is the oldest gcc I want to support.

zlib in Emscripten should be as simple as adding `-s USE_ZLIB=1`. Anyway, it is only used to compress tree data for 3D honeycombs (honeycomb-rules files) which are currently disabled in the Emscripten version by default, so I have made the Emscripten build not depend on ZLIB.

PR #92, opened 2020-02-23
Add `-pedantic` to the Makefile, and fix all resulting warnings

Quuxplusone:
The individual commits here are all severable, but I recommend taking all of them anyway.

"Add a helper function `voidp` to silence GCC warnings about `%p`" is extremely awkward, but unfortunately it has to be done, if we want to have both `-Wformat` and `-pedantic`. [GCC doesn't provide an equivalent of Clang's `-Wno-format-pedantic` option.](https://stackoverflow.com/questions/27975147/gcc-switch-or-pragma-to-suppress-the-p-format-warning-for-non-void-pointers)

"Lambdas can't have default arguments until C++14" is mildly unfortunate: The code you wrote is perfectly valid C++14, but we currently compile as C++11, where `-pedantic` flags it as an error. Upgrading to C++14 would be another option; Clang 3.5 and GCC 5.1 both support default arguments in lambdas with `-std=c++14`.


Quuxplusone:
Ping! Rebased on 11.3.

Quuxplusone:
@zenorogue gentle ping!

zenorogue:
Merged. Thanks!

PR #101, opened 2020-03-31
langen: Handle repeated strings better.

Quuxplusone:
If a translation file contains two copies of the same English string,
with the same translation, then we don't care at all; just throw out
the second copy.

If a translation file contains two copies of the same English string,
with DIFFERENT translations, then (according to how the game does
translation) one of those translations must be inappropriate. Give
a warning in that case; and again, throw out the second copy because
we are forced to pick ONE translation for use within the game.

Also fix some minor typos I noticed in the strings.


Quuxplusone:
@zenorogue: I found the Red Raiders typo organically. Then I wrote a Python script to find other places where the translations had strings that didn't quite match the English strings. You can see what I did here — https://github.com/Quuxplusone/hyperrogue/commit/translations-bonus — but I definitely don't expect you to put anything like that into the repo.

There are tons of obsolete strings in `language-??.cpp` that no longer appear in the program itself. Removing these strings makes `language-data.cpp` 8% smaller (from 1831KB to 1687KB), and makes `./hyperrogue` itself 1% smaller (from 9603KB to 9512KB) on my platform.

Quuxplusone:
@zenorogue ping!

PR #105, opened 2020-04-12
Make "mymake" work on OSX, and add it to Travis

Quuxplusone:
```
commit a49c40cee768128a40223fcb74165a595109c220
Author: Arthur O'Dwyer <arthur.j.odwyer@gmail.com>
Date:   Fri Apr 3 22:01:26 2020 -0400

    Make "mymake" work on OSX, and add it to Travis.
    
    Fix a -Wformat bug exposed by compiling with Clang.
    
    To preprocess C++11 code, you need `g++ -E -std=c++11`, not just
    `g++ -E`. (The old code worked for GCC 6+ and Clang 6+ because they
    changed the default mode from C++03 to C++14. But for GCC 5, we still
    need `-std=c++11`. And regardless, it's a good idea.)
    
    Add a "-mac" option to mymake, and cleanly factor out `set_mac`,
    `set_linux`, and `set_win`. When you build mymake using
    `make -f Makefile.simple mymake`, you get a mymake that knows what
    platform it's on. This means you don't have to pass `mymake -mac`
    on OSX, nor `mymake -win` on Windows.
    
    The old code put `INCLUDE(___hyper-main.cpp)` into a C++ file that
    would be preprocessed, which doesn't work because libSDL does
    essentially `-Dmain=SDL_main`, which turns this into
    `INCLUDE(___hyper-SDL_main.cpp)`, which gives us
    a "file not found" error from mymake. The solution is to put
    filenames into quotation marks, so that the string "main" never
    appears as a token in the C++ file. (Alternatively, we could have
    renamed "hyper-main.cpp" to "hypermain.cpp".)
    
    Add several new "mymake" entries in the Travis build matrix,
    and add the "mymake" builds' badge to the README.


commit 6e4b6ee93cf907fff0b85bc8b9ddd11d3b11333a
Author: Arthur O'Dwyer <arthur.j.odwyer@gmail.com>
Date:   Sat Apr 4 15:53:08 2020 -0400

    Remove the definition of _Unwind_Resume on OSX.
    
    This is unnecessary for me when using Makefile.simple, as far as I can tell.
    And it's actively harmful when using "mymake", because then we get this
    non-inline variable defined in every .o file, which leads to
    multiple-definition errors at link time.
    
    Maybe it's no longer needed?
```


Quuxplusone:
@zenorogue ping!

zenorogue:
Thanks!

I will look into whether _Unwind_Resume is still needed on my machine, if yes, I will add it again in a better location.

It appears that set_win() is currently rather useless (it was supposed to compile the Windows version via Wine using my batch files, IIRC it did not actually work, and for other people who don't have these files it is even more useless). If you want to fix this, please do.

Quuxplusone:
> It appears that set_win() is currently rather useless ... If you want to fix this, please do.

I don't have access to a Windows machine; @danfun64 is the one who usually reports Windows issues. :) I spent some time a while back trying to get HyperRogue to compile on [Appveyor](https://www.appveyor.com). TravisCI now also has a "Windows" configuration available, but I never investigated it far enough to learn how to install all the dependencies HyperRogue needs (SDL, libpng, etc).

PR #108, opened 2020-05-03
Move all defaulting-of-`CAP_FOO` to sysconfig.h

Quuxplusone:
- The phrase `#ifdef CAP_` should never appear anywhere.

- The phrase `#ifndef CAP_` should appear only in sysconfig.h.

- The phrases `#if CAP_` and `#if !CAP_` may appear wherever,
    as long as "sysconfig.h" is included first.

The rules for `CAP_FOO` equally apply to `ISFOO`.

There are many one-off macros still tested with `#ifdef`,
including `HAVE_ACHIEVEMENTS`, `PRINT_ACHIEVEMENTS`,
`FAKEWEB`, `FAKE_SDL`, `EASY`, and `WHATEVER`. I don't
have much grasp on what these are used for or how they're
configured, so I'm leaving them alone.

### DIFF
--- a/achievement.cpp
+++ b/achievement.cpp
@@ -952,12 +952,12 @@ EX void achievement_display() {
     col /= 10; col *= 0x10101;
     displayfr(vid.xres/2, vid.yres/4, 2, vid.fsize * 2, achievementMessage[0], col & 0xFFFF00, 8);
     int w = 2 * vid.fsize;
-#if ISMOBILE==0
+#if !ISMOBILE
     while(w>3 && textwidth(w, achievementMessage[1]) > vid.xres) w--;
 #endif
     displayfr(vid.xres/2, vid.yres/4 + vid.fsize*2, 2, w, achievementMessage[1], col, 8);
     w = vid.fsize;
-#if ISMOBILE==0
+#if !ISMOBILE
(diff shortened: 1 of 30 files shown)


Quuxplusone:
@zenorogue gentle ping!
Particularly commit f56e0fe2e3ef2dc315f51b7fa58722e99396ceb6 would fix a lot of the red on the buildbot right now.

zenorogue:
Thanks!

One small thing I did not agree with: CAP_NCONF, CAP_RVSLIDES and CAP_DRAW are RogueViz-specific flags, so I think HyperRogue should not know about them (i.e., they should not appear in sysconfig.h). I have moved CAP_NCONF and CAP_RVSLIDES to rogueviz.h, and CAP_DRAW to newconf.cpp (as it is actually newconformist's flag used to configure it, not a HyperRogue/RogueViz flag).

Quuxplusone:
Ah, newconformist being https://github.com/zenorogue/newconformist , right? I wasn't previously aware of it. I guess that makes sense. FWIW, I don't understand the intended behavioral difference between "newconf.cpp"-with-`CAP_NCONF` and "newconf.cpp"-without-`CAP_NCONF`.

PR #111, opened 2020-05-08
Replace "%The1 is confused!" with "You have inexplicably cheated death!"

### LINKED ISSUE #95: Sandworm can path through boats, and "%The1 is confused!"
I don't fully understand what's going on here — I'm as confused as the poor sandworm is!

If you lead a sandworm into the Living Fjord, and then you get in a boat, you can be next to the sandworm's head and then step sideways (as if you were killing a Hedgehog Warrior), and the sandworm will wind up in the same cell as you. This seems to work only if you are in a boat. On land, "the sandworm would kill you there."

You can step off onto land: 

Or, you can step off into water, which will move your boat, leaving the sandworm floating in the water. The sandworm will then explode on your next turn, _unless_ you move your boat hedgehog-wise and allow the worm to step under you again. By repeatedly doing this, you can move the sandworm across a river using only one boat: 

Or, if you have a lot of boats, the sandworm will happily path across them: 

Finally, if instead of stepping off, you click your own cell instead, you get the message "You need to dismount the Sand Worm W!" 

That is, the game thinks you're "riding" the sandworm, simply because it shares your cell. However, from the code, it seems like the only _intended_ way to ride a sandworm is to use the Orb of Domination, right?

My suggested fix for the immediate issue is that sandworms should never path through boats at all; if you're in a boat, you should be safe from sandworms. However, it seems like there must be some other issue that causes the sandworm to move onto the player's own cell, in general, that might also be triggered by something other than boats. (I would guess that I could trigger the same bug in R'Lyeh by moving statues? I haven't tested that yet.)

Separately, I think it's strange that the message is "You need to dismount the Sand Worm <b>W</b>!" That "W" seems out of place. I can't figure out where it's coming from.

Separately, I've looked in the code at where the message `"%The1 is confused!"` is printed, and I think the message should be changed to `"You have inexplicably cheated death!"` or something along those lines. The current message implies that I've done something clever to the sandworm in-universe; but in the code, it's apparent that this is basically an assertion-failure and the whole game is in a bad state _out-of-universe_.

Quuxplusone:
> The current message implies that I've done something clever
> to the sandworm in-universe; but in the code, it's apparent
> that this is basically an assertion-failure and the whole game
> is in a bad state *out-of-universe.*

Fixes #95.


Quuxplusone:
@zenorogue ping?

Quuxplusone:
@zenorogue ping!

zenorogue:
There are some other controversial places where the game texts do not exactly tell the truth, like the description of the Dead Orb (which claims it is useless) and the description of the Palace (which says you will never find the Prince(ss), which was true when Palace was originally implemented, but no longer).

I like them myself, they express what the character/narrator is thinking when they find these things, even if they are slightly wrong.

Quuxplusone:
> they express what the character/narrator is thinking when they find these things

Yeah, but this particular message doesn't, right? Because "what the narrator is thinking (about the state of mind of the sandworm)" is an _in-universe_ thing; but this message is talking about the game itself getting into a bad state _out-of-universe_. Arguably it would be okay for the message to refer to the player-character's own mental state — `"You are very confused about what just happened!"` — but I continue to think that attributing confusion to the _monster_ in this case is just wrong.

jruderman:
How about "You somehow avoid being killed by %the1"

* Staying away from "inexplicably" and "cheated" reduces the suggestion that the player has gained an unfair advantage by cheating.
* Keeps the monster name in the message, which is important for working out steps-to-reproduce

Quuxplusone:
How about "You unexpectedly avoid being killed by %the1!"
or "Surprisingly, %the1 does not kill you!"

I see your objection to "cheated [death]" (especially for a game whose players, I assume, often aren't native speakers of English), but I'm not so sure I understand the objection to "inexplicably." I would like to keep some element of "This is an unexpected, surprising, inexplicable situation!" as just a minor hint that what's just happened is essentially a transient bug in the game logic, not an intentional in-game effect that you should expect to rely on.

ammongit:
I think my biggest issue with the replacement message is it doesn't quite work if you're not actually in danger. If a sandworm off in the distance gets bugged and prints the "confused" message, I'm not exactly "cheating death" there.

Quuxplusone:
> I think my biggest issue with the replacement message is it doesn't quite work if you're not actually in danger. If a sandworm off in the distance gets bugged and prints the "confused" message, I'm not exactly "cheating death" there.

Given that the message is printed only on an assertion failure inside `killThePlayer`, I don't think it's possible to get the message due to anything happening in the _distance_. It's also not specific to sandworms. I claim that your interpretation of the existing message as "oh, something interesting happened to the mental state of _that sandworm_" is exactly the kind of player confusion that I'd like to resolve by rewording the message. :)

ammongit:
I didn't realize this was part of `killThePlayer`, I thought it was a general replacement for that message.

PR #112, opened 2020-05-16
Some fixes to mymake and Makefile.simple

rpljwi:
Just a simple override in case there is a `-jN` parameter in `MAKEFLAGS`. This seems necessary, because otherwise `./makeh` could be executed before `./langen` despite the former needing the latter's output.

### DIFF
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -94,7 +94,7 @@ int main(int argc, char **argv) {
       obj_dir += "/";
       setdir += "../";
       for(char c: s) 
-        if(c == '=' || c == '-' || c == '/') obj_dir += "_"; 
+        if(c == '=' || c == '-' || c == '/' || c == '"' || c == '\\') obj_dir += "_"; 
         else obj_dir += c;      
       }
     else if(s == "-win") {
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -83,8 +83,10 @@ int main(int argc, char **argv) {
   set_linux();
 #endif
   for(string fname: {"Makefile.loc", "Makefile.simple", "Makefile"})
-    if(file_exists(fname))
+    if(file_exists(fname)) {
       system("make -f " + fname + " language-data.cpp autohdr.h savepng.o");
+      break;
+      }
   for(int i=1; i<argc; i++) {
     string s = argv[i];
     if(s.substr(0, 2) == "-D") {
--- a/Makefile.simple
+++ b/Makefile.simple
@@ -152,7 +152,7 @@ langen$(EXE_EXTENSION): langen.cpp language-??.cpp language-ptbr.cpp
 makeh$(EXE_EXTENSION): makeh.cpp
 	$(CXX) -O2 makeh.cpp -o $@
 
-autohdr.h: makeh$(EXE_EXTENSION) *.cpp
+autohdr.h: makeh$(EXE_EXTENSION) language-data.cpp *.cpp
 	./makeh classes.cpp locations.cpp hyperpoint.cpp geometry.cpp goldberg.cpp init.cpp floorshapes.cpp cell.cpp multi.cpp shmup.cpp pattern2.cpp mapeditor.cpp graph.cpp textures.cpp hprint.cpp language.cpp util.cpp complex.cpp *.cpp > autohdr.h
 
 language-data.cpp: langen$(EXE_EXTENSION)


Quuxplusone:
Surely it would be better to add `language-data.cpp` as a dependency of `autohdr.h`.
The production right now is

```
autohdr.h: makeh$(EXE_EXTENSION) *.cpp
        ./makeh classes.cpp locations.cpp hyperpoint.cpp geometry.cpp goldberg.cpp init.cpp floorshapes.cpp cell.cpp multi.cpp shmup.cpp pattern2.cpp mapeditor.cpp graph.cpp textures.cpp hprint.cpp language.cpp util.cpp complex.cpp *.cpp > autohdr.h
```

I don't know why makeh wants to be invoked with those specific 18 .cpp files _followed by_ all of the .cpp files; but anyway, it should definitely depend on "language-data.cpp" if that's expected to be one of the inputs.

rpljwi:
Don't think it matters too much from perspective of compilation time, since mymake seemingly compiles most of the files in a serial fashion, but yeah, that's likely a better way to go around. Moreover, it looks like Makefile.am already does it this way: https://github.com/zenorogue/hyperrogue/blob/4cee27ae767fcb60daf629c7a5bdced62bdfc621/Makefile.am#L28

rpljwi:
It seemed related enough, so here's something else.
https://github.com/zenorogue/hyperrogue/blob/4cee27ae767fcb60daf629c7a5bdced62bdfc621/mymake.cpp#L85-L87
This seems odd in case there actually is more than one file from the list. Don't know the original intention, but probably the idea is that there can only be one.

Quuxplusone:
I think you're right that the list is ordered by priority. `Makefile.loc` ("loc" for "local"?) does not exist in the repo, but my wild guess is that @zenorogue sometimes drops one into the current directory just to test something out. `Makefile.simple` is the "official" makefile, such as it is. A file named `Makefile` is generated by automake, if you use the automake setup (but I am not aware that anyone is actively attempting to maintain the automake setup).

I agree, it seems like it would be cool if this loop could be simplified down to the single line

    system("make -f Makefile.simple language-data.cpp autohdr.h savepng.o");

rpljwi:
Yeah, about that automake approach... I had a problem where ./configure had been unable to locate SDL_gfxPrimitives.h despite the file clearly being present. ...But that's another story altogether.

rpljwi:
There is another thing that could need fixing in `mymake.cpp`, related to escaping some input strings. Forgive my ignorance (not an experienced developer by any means), but should I just push to this PR, or rather open a new one? It's just a change in one line.

Quuxplusone:
@rpljwi: Obviously I'm not the authority on this particular repo :) but in general I'd say that if the change is entirely orthogonal to this one, logically it should go in a separate PR. But I admit I've made plenty of "omnibus bill" PRs against this repo myself; e.g. #92.

Another note on PRs in general (without meaning to imply any special authority over _this_ repo): In this PR right now you've got two commits, one adding `-j1` and the other reverting `-j1`-and-adding-the-dependency. IMO you should squash them into a single commit so that the `-j1` stuff no longer appears in the commit history. You can do that with a sequence such as

    git checkout make-j1
    git rebase -i HEAD~~
        ("pick" the first commit and "squash" the second; edit the commit message and save it)
    git push -f origin make-j1

rpljwi:
Thanks for advice.
Yup, I thought about squashing, but wouldn't the discussion then seem kinda without context, what with commit history changed? Then again, it's not like these changes in particular are super-complex, so it's OK I guess.
And the other changes are not entirely orthogonal either, and they're all tiny... so probably just will have to change PR name again.

rpljwi:
The first commit was already discussed. The second is about multiple `Makefile`s and was discussed too. The last one needs some explanation.
Some `#define`s allow specifying path prefixes for file lookup, e.g. `HYPERPATH` is prepended to paths to font, music and sounds. So it seems worth to be able to pass that kind of argument to `mymake`, but the resulting hierarchy of escaping needs some attention.

mymake.cpp
```diff
@@ -83,16 +83,18 @@ int main(int argc, char **argv) {
...
+      }
   for(int i=1; i<argc; i++) {
     string s = argv[i];
     if(s.substr(0, 2) == "-D") {
       opts += " " + s;
       obj_dir += "/";
       setdir += "../";
       for(char c: s) 
-        if(c == '=' || c == '-' || c == '/') obj_dir += "_"; 
+        if(c == '=' || c == '-' || c == '/' || c == '"' || c == '\\') obj_dir += "_";
```
Quuxplusone, line 97: The only really dangerous character on POSIX is `/`; but the shell might care about any of `'"$!()[]{};%*&<>`. To be safe I'd just replace these two lines with `obj_dir += (isalnum(c) ? c : '_');`

rpljwi: Makes sense. Probably better to replace more "problematic" characters, since there is also Windows, for example, under which I'm currently trying to coerce `mymake` into working.
There is also an `if` to pretty much the same effect here:
https://github.com/zenorogue/hyperrogue/blob/0813fb65ca565cadf7baa971615a643e06be1357/mymake.cpp#L120
I guess the same reasoning would apply there? And at any rate, now that @zenorogue has merged the PR, what would be the appropriate thing to do to make that change?

zenorogue: I have pushed the change where I replaced the check with `!isalnum(c)` like @Quuxplusone suggested. I think making mymake work on Windows should be another PR. Thanks!

rpljwi: Yeah, I'll create a PR if I succeed (has not yet happened). OK, will now delete the branch then.

PR #113, opened 2020-05-17
Add basic highlighting of enemies on the horizon

still-flow:
So the cells at the edge of the map (at least in the usual Poincare model) are tiny, making it kind of hard to notice monsters there. Of course, Alt-highlight and high contrast modes already exist, but, at least for me, they didn't help all that much. Both suffer from the fact that the highlighted shape is still tiny. Also, they are static.

Cue one of the simplest possible solutions: use "special aura" with an oscillating color to draw attention to monsters you could otherwise overlook. One possible caveat is that other things also use it, so there may be conflicts. I'm not even completely sure that drawMonster(...) is the right place for that.

In all, the code with this PR is probably not all that important (it's likely ignoring many non-standard cases anyway). Obviously there is more than one way to approach the outlined problem, so I hope it should at least stimulate the discussion.

### DIFF
--- a/config.cpp
+++ b/config.cpp
@@ -340,6 +340,7 @@ EX void initConfig() {
   addsaver(pconf.ballproj, "ballproj", 1);
   addsaver(vid.monmode, "monster display mode", DEFAULT_MONMODE);
   addsaver(vid.wallmode, "wall display mode", DEFAULT_WALLMODE);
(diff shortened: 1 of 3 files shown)


zenorogue:
Thanks! Improved this a bit (configurable threat level, configurable color) and pushed.

Issue #107, opened 2020-04-19
No longer builds when CAP_PNG=0


Quuxplusone:
```
$ make -f Makefile.simple 
./langen > language-data.cpp
./makeh classes.cpp locations.cpp hyperpoint.cpp geometry.cpp goldberg.cpp init.cpp floorshapes.cpp cell.cpp multi.cpp shmup.cpp pattern2.cpp mapeditor.cpp graph.cpp textures.cpp hprint.cpp language.cpp util.cpp complex.cpp *.cpp > autohdr.h
c++ -O2 -std=c++11 -DMAC -I/usr/local/include -march=native -W -Wall -Wextra -Werror -pedantic -Wno-unused-parameter -Wno-implicit-fallthrough -Wno-maybe-uninitialized -Wno-unknown-warning-option -DCAP_GLEW=0 -DCAP_PNG=0    -c hyper.cpp -o hyper.o
In file included from hyper.cpp:83:
./mapeditor.cpp:2330:26: error: use of undeclared identifier 'texture'
          if(!holdmouse) texture::config.data.undoLock();
                         ^
```

This happens on Mac if you don't pass `HYPERROGUE_USE_PNG=1`. Because without `CAP_PNG`, there's no `CAP_TEXTURE`, and without `CAP_TEXTURE` there's no `namespace texture`.

The TravisCI matrix does not test `CAP_PNG=0`. I would be happy to submit a PR to add a couple `CAP_PNG=0` builds to the test matrix; alternatively, I would be happy to submit a PR to make `-DCAP_PNG=1` mandatory and eliminate all the conditionals that depend on it.

Quuxplusone:
Orthogonal but minor: It would be nice to add a commit hook, or some check in `makeh`, or something, that would fail the build whenever it saw `#ifdef CAP_FOO` anywhere, or `#ifndef CAP_FOO` outside of "sysconfig.h". The directives `#if CAP_FOO` and `#ifdef CAP_FOO` do different things when `-DCAP_FOO=0`.

```
diff --git a/complex2.cpp b/complex2.cpp
index e861c827..11c490b2 100644
--- a/complex2.cpp
+++ b/complex2.cpp
@@ -8,7 +8,7 @@
  */
 
 #include "hyper.h"
-#ifdef CAP_COMPLEX2
+#if CAP_COMPLEX2
 
 namespace hr {
 ```

Quuxplusone:
[...]

PR #115, opened 2020-05-20
Attempt to convince mymake to work on MinGW

still-flow:
...has not been as successful as I'd hoped. Short version: build runs just fine, but program segfaults upon launch.

Setup: MinGW64 under MSYS2 of latest version as of writing. Necessary packages installed (btw, I could probably try to make a config for Travis with this setup; looking at Travis docs it didn't seem too difficult).
Preliminary note: builds with `automake` and `Makefile.simple` both work fine. At least no segfault at launch.

A bit about proposed changes to `mymake.cpp`. I based the configuration off of `Makefile.simple`. Options such as `-DCAP_PNG` are just temporarily there, as per #107.
As for renaming `win` to `mingw64`, this was to be more specific _just_ in case there are other Windows build profiles. Also, `system(cmd)` on MinGW is peculiar and required a workaround. It's really only needed for `mkdir` and `windres`, but I don't think it hurts everywhere else.

And now, the segfaults. I tried different permutations and combinations of options, but the result is the same: `hyper.exe`, if ran from MinGW shell, segfaults straight away. Not even `--help` option works. And ran natively from Windows (double-clicking the file, that is) it opens a console window, where usually the various debug info would be printed, prints nothing, hangs for a second, and closes.

Running from under MinGW, `gdb` shows that segfaults occur in different functions sometimes, but all of them have "dunder" names, which AFAIK means that these are a part of the C++ runtime. I'll post backtraces (and whatever else is required) in a bit. Admittedly, I don't know all that much about MinGW or C++ in general, so apologies for not knowing what info exactly to provide.

With that said, I came across something called "static initialization order fiasco", but I'm not sure it's relevant. One of the segfault cases did involve a function `__static_initialization_and_destruction_0(int, int)`, though.

And that's all for now, I think.

### DIFF
--- a/.github/workflows/test_simple.sh
+++ b/.github/workflows/test_simple.sh
@@ -6,6 +6,7 @@ if [[ "$GH_OS" == "windows-latest" && "$GH_BUILDSYS" == "mymake" ]]; then
 cat << ENDOFCMDS > .github/workflows/gdb_cmds.txt
   run --version
   backtrace
+  exit 1
 ENDOFCMDS
 
   gdb --batch -x .github/workflows/gdb_cmds.txt ./hyperrogue
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -32,6 +32,7 @@ string linker;
 string libs;
 
 int batch_size = thread::hardware_concurrency() + 1;
+bool mingw64 = false;
 
 void set_linux() {
   preprocessor = "g++ -E";
@@ -49,14 +50,14 @@ void set_mac() {
   libs = " savepng.o -L/usr/local/lib -framework AppKit -framework OpenGL -lSDL -lSDLMain -lSDL_gfx -lSDL_mixer -lSDL_ttf -lpng -lpthread -lz";
   }
 
-void set_win() {
+void set_mingw64() {
+  mingw64 = true;
   preprocessor = "g++ -E";
-  compiler = "runbat bwin-g.bat -c";
-  linker = "runbat bwin-linker.bat";
-  opts = "-DFHS -DLINUX -I/usr/include/SDL";
-  libs = "";
-
-  standard = "";
+  compiler = "g++ -mwindows -march=native -W -Wall -Wextra -Werror -Wno-unused-parameter -Wno-implicit-fallthrough -Wno-maybe-uninitialized -c";
+  linker = "g++ -o hyper";
+  opts = "-DWINDOWS -DCAP_GLEW=1 -DCAP_PNG=1";
+  libs = " savepng.o hyper.res -lopengl32 -lSDL -lSDL_gfx -lSDL_mixer -lSDL_ttf -lpthread -lz -lglew32 -lpng";
+  setvbuf(stdout, NULL, _IONBF, 0); // MinGW is quirky with output buffering
   }
 
 vector<string> modules;
@@ -73,6 +74,8 @@ string obj_dir = "mymake_files";
 string setdir = "../";
 
 int system(string cmdline) {
+  if (mingw64)
+    cmdline = "sh -c '" + cmdline + "'"; // because system(arg) passes arg to cmd.exe on MinGW
   return system(cmdline.c_str());
   }
 
@@ -84,7 +87,7 @@ int main(int argc, char **argv) {
 #if defined(MAC)
   set_mac();
 #elif defined(WINDOWS)
-  set_win();
+  set_mingw64();
 #else
   set_linux();
 #endif
@@ -103,9 +106,9 @@ int main(int argc, char **argv) {
         if(!isalnum(c)) obj_dir += "_"; 
         else obj_dir += c;      
       }
-    else if(s == "-win") {
-      set_win();
-      obj_dir += "/win";
+    else if(s == "-mingw64") {
+      set_mingw64();
+      obj_dir += "/mingw64";
       setdir += "../";
       }
     else if(s == "-mac") {
@@ -267,7 +270,9 @@ int main(int argc, char **argv) {
       }
     else if (tasks_done == tasks_amt) { finished = true; break; }
     } this_thread::sleep_for(quantum); }
-  
+
+  if (mingw64) (void)system("windres hyper.rc -O coff -o hyper.res");
+
   printf("linking...\n");
   system(linker + allobj + libs);
   return 0;
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -91,9 +91,11 @@ int main(int argc, char **argv) {
 #else
   set_linux();
 #endif
+  int retval = 0; // for storing return values of some function calls
   for(string fname: {"Makefile.loc", "Makefile.simple", "Makefile"})
     if(file_exists(fname)) {
-      system("make -f " + fname + " language-data.cpp autohdr.h savepng.o");
+      retval = system("make -f " + fname + " language-data.cpp autohdr.h savepng.o");
+      if (retval) { printf("error during preparation!\n"); exit(retval); }
       break;
       }
   for(int i=1; i<argc; i++) {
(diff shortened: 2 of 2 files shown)


rpljwi:
Oh, and there is _something_ wrong with using `make -f Makefile.simple mymake`. That causes MinGW to open a different cmd.exe window every time `system(cmd)` is invoked. Probably warrants a separate issue/PR though. For now, `make mymake` and passing `-mingw` to `mymake` works.

rpljwi:
Here's one backtrace: https://gist.github.com/rpljwi/c5937a5397e2ab8ce63955e64832e02f

By the way, program needs several DLLs, which for now I just have to copy into the main directory. There must be a better solution (could that cause segfaults?).

Quuxplusone:
> btw, I could probably try to make a config for Travis with this setup

That would be awesome!

> That causes MinGW to open a different cmd.exe window every time system(cmd) is invoked, and that consequently makes everything misbehave

I don't understand the issue here. The number of windows shouldn't matter as long as the specified command actually gets run. Is it something about `system` running the specified command in the wrong current-working-directory?

> https://gist.github.com/rpljwi/c5937a5397e2ab8ce63955e64832e02f

Weird; initializing a `constexpr hyperpoint` from another (trivially copyable) `hyperpoint` shouldn't be able to segfault. Maybe the backtrace is "wrong" because of inlining? If you replace `-O2` with `-O0` on all the compiler command lines, what backtrace do you get?
Just to check my premises: if you add `static_assert(std::is_trivially_copyable<hyperpoint>::value, "");` somewhere around there, it doesn't assert-fail at compile time, does it?
I do think it's a bad idea that `hyperpoint` inherits from `std::array`, but I can't really see how that would cause this symptom.

rpljwi:
> I don't understand the issue here. The number of windows shouldn't matter as long as the specified command actually gets run. Is it something about `system` running the specified command in the wrong current-working-directory?

For one, it nullifies the workaround for `mkdir`. Windows-native `mkdir` doesn't know `-p` argument, instead creating a directory named `-p`. But it's possible to call `sh` since `cmd.exe` is invoked with PATH set up correctly, I guess. It's strange since in "multiple-windows" case `g++` is invoked just fine, but a directory named `-p` still appears.

> If you replace `-O2` with `-O0` on all the compiler command lines, what backtrace do you get?

> Just to check my premises: if you add `static_assert(std::is_trivially_copyable<hyperpoint>::value, "");` somewhere around there, it doesn't assert-fail at compile time, does it?

I'll try all of that.

rpljwi:
Bizarre. I replaced every `-O2` with `-O0`, even in Makefile.simple, and now I get no progress report updates, despite object files being created. Optimizations affecting `printf`?

rpljwi:
So, `-O0` didn't seem to change the backtrace at all. Compared to the one posted, only thread ids were different. But the caveat is stated in the edit above: I'm doubting my MinGW installation now.

rpljwi:
Modifying `hypgraph.cpp` as follows
```
static_assert(std::is_trivially_copyable<hyperpoint>::value, "");
hyperpoint ghxy, ghgxy, ghpm = C0;
```
didn't fail the build process.

rpljwi:
Haven't made any progress on determining the cause, but managed to [reproduce](https://github.com/rpljwi/hyperrogue/runs/702475206) the same backtrace with Github's CI.

zenorogue:
`C0` is a macro:

```
#define C0 (MDIM == 3 ? C02 : C03)
#define MDIM (MAXMDIM == 3 ? 3 : cginf.g.homogeneous_dimension)
#define cginf ginf[geometry]
vector<geometryinfo> ginf = { ... }
```

so it seems that `C0` is initialized before `ginf`.

I have changed `C0` to `C02`, does that fix the problem? (There might be some other global variable initialized to `C0` somewhere.)

still-flow:
> I have changed `C0` to `C02`, does that fix the problem?

I've ported this PR's changeset onto the recent master, and no luck. Still the same old segmentation fault, albeit [in a different spot](https://gist.github.com/still-flow/fa78cb7deecaee28c7897fe7b0532a51).

I feel I'm missing something fundamental regarding how linking on Windows is different from Unices, but this is just a hunch stemming from my ignorance of the details of the linking process in general.

(Unfortunately, there currently seems to be a packaging mishap with the MinGW64 on Github's runner leading to run failure during installation of dependencies. Not that it matters in this case since no config for mymake on Windows at present.)

zenorogue:
Initialization order fiasco again -- if X is initialized in file A and Y is initialized in file B, the linker does not know which one should be initialized first. If X depends on Y and the linker decides to initialize X first, there is a problem.

ringcolor was initialized as `darkena(0xFF, 0, 0xFF)` which made no sense and also depended on GDIM (which has the same issue as MDIM). Changed to `ringcolor = 0xFFFF` which should fix this.

still-flow:
[Segfault in a new place again!](https://gist.github.com/still-flow/ccb254baf546269e01d8bff2568341d7)

I think at this point it would be more helpful if there was an entry in the CI build matrix for this, so I'm gonna create a PR adding it. Otherwise we might be going back and forth like that for a long time :).

zenorogue:
This one should be fixed. I am more optimistic, and hope this will end soon :)

still-flow:
So [this error](https://github.com/zenorogue/hyperrogue/runs/1160043186?check_suite_focus=true#step:4:35) happened. Since when did *this* become a thing? Probably should cast that `system` call to `void`, but it's baffling that it's decided to break now. New `gcc` update with new defaults maybe?

Quuxplusone:
> `error: ignoring return value of ‘int system(const char*)’, declared with attribute warn_unused_result [-Werror=unused-result]`

This usually isn't a sign of a _new_ GCC but rather a pretty old one. glibc used to be really really anal about `warn_unused_result` — putting it on, like, `write` and `fscanf` — and then radically scaled it back a few years ago. The only standard function that triggers `-Wunused-result` in a modern glibc, AFAIK, is `realloc`.

But I see absolutely nothing wrong with adding a `(void)` cast to shut up the warning here.

still-flow:
> But I see absolutely nothing wrong with adding a `(void)` cast to shut up the warning here.

I mean, it's not hard to do, but _why_ does only this call trip it up? There are like two dangling `system` calls in other places, and these are somehow fine? Is that because this one is in a body of an `if`-statement?

still-flow:
Well, scopes be damned.

@zenorogue: I have a horrible idea. Since you're feeling optimistic about the source of the segfaults, how about merging this -- provided that no-one finds any glaring flaws -- so that you could iterate on them at your own pace? I wish I could help with that, but looking at your other fixes I see that they required some grasp of what each global variable is supposed to do, which at least currently I'm afraid is above my capabilities.

I still don't understand why everything worked just fine on MacOS and Ubuntu, though.

Quuxplusone:
@still-flow: SGTM. But I still don't understand how to navigate the GitHub Workflows CI builds. I'm looking at https://github.com/zenorogue/hyperrogue/pull/115/checks?check_run_id=1161955292 and I see three builds failed with red Xs in the left pane. When I click on the first two (GCC failures) it takes me directly to a compiler error message, which is useful. But when I click on the third (Clang failure) it takes me to a thing (depicted in the screenshot below) that says "Build" failed, but doesn't provide any information about why or how.



How do I see the failure in this third case?

still-flow:
> How do I see the failure in this third case?

I think that one timed out for some reason -- see that its time is about 30 minutes, much larger than usual? I agree that it's no good that there is nothing log-wise indicating the reason though. Wild guess: something hanged at the runner service level (or whatever the level above the virtual environment in which builds run is called), hence no logs.

still-flow:
Modified the GDB batch files so that Windows-mymake tests fail for now -- for convenience's sake, since clicking on the failed check automatically opens the log in the relevant place.

PR #116, opened 2020-05-20
Add Travis configuration for MinGW64 builds

rpljwi:
Based on: https://docs.travis-ci.com/user/reference/windows/#how-do-i-use-msys2

Just one -- Makefile.simple -- job for now, since that seemed to work fine locally for me. Automake-build could also be added, but not sure of its necessity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,6 +132,22 @@ before_install:
     brew install sdl sdl_gfx sdl_mixer sdl_ttf
     # work around https://stackoverflow.com/questions/51034399/ for now
     (cd /usr/local/include && ln -sf SDL/SDL.h)
+  elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
+    [[ ! -f C:/tools/msys64/msys2_shell.cmd ]] && rm -rf C:/tools/msys64
+    choco uninstall -y mingw
+    choco upgrade --no-progress -y msys2
+    export msys2='cmd //C RefreshEnv.cmd '
+    export msys2+='& set MSYS=winsymlinks:nativestrict '
+    export msys2+='& C:\\tools\\msys64\\msys2_shell.cmd -defterm -no-start'
+    export mingw64="$msys2 -mingw64 -full-path -here -c "\"\$@"\" --"
+    export msys2+=" -msys2 -c "\"\$@"\" --"
+    $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-toolchain
+    ## Install more MSYS2 packages from https://packages.msys2.org/base here
+    $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-SDL mingw-w64-x86_64-SDL_gfx mingw-w64-x86_64-SDL_mixer mingw-w64-x86_64-SDL_ttf
+    taskkill //IM gpg-agent.exe //F  # https://travis-ci.community/t/4967
+    taskkill //IM dirmngr.exe //F # this process also seemed to hang after install
+    export PATH=/C/tools/msys64/mingw64/bin:$PATH
+    export MAKE=mingw32-make  # so that Autotools can find it
   else
     exit 'Unsupported OS'
   fi
@@ -142,6 +158,8 @@ before_install:
       sudo apt-get install -qq libglew-dev
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew install glew
+    elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
+      $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-glew
     else
       exit 'Unsupported OS'
     fi
@@ -153,6 +171,8 @@ before_install:
       echo 'libpng is installed by default'
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew install libpng
+    elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
+      $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-libpng
     else
       exit 'Unsupported OS'
     fi
@@ -164,6 +184,8 @@ before_install:
       echo 'autotools is installed by default'
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew install automake
+    elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
+      $msys2 pacman --sync --noconfirm --needed automake-wrapper
     else
       exit 'Unsupported OS'
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,6 +120,15 @@ matrix:
       HYPERROGUE_USE_GLEW=1
       HYPERROGUE_USE_PNG=1
       HYPERROGUE_USE_ROGUEVIZ=1
+  - os: windows
+    compiler: gcc
+    env: >-
+      TRAVIS_OS_NAME=windows
+      TRAVIS_COMPILER_NAME=gcc
+      TRAVIS_BUILD_SYSTEM=Makefile
+      HYPERROGUE_USE_GLEW=1
+      HYPERROGUE_USE_PNG=1
+      HYPERROGUE_USE_ROGUEVIZ=1
 
 before_install:
 - |-
@@ -199,7 +208,7 @@ script:
     ./configure CXXFLAGS="-W -Wall -Wextra -Werror -Wno-unused-parameter -Wno-maybe-uninitialized -Wno-unknown-warning-option"
     make
   elif [[ "$TRAVIS_BUILD_SYSTEM" == "Makefile" ]]; then
-    make -f Makefile.simple
+    $mingw64 make -f Makefile.simple
   elif [[ "$TRAVIS_BUILD_SYSTEM" == "mymake" ]]; then
     make -f Makefile.simple mymake
     if [[ "$HYPERROGUE_USE_ROGUEVIZ" == "1" ]]; then


.travis.yml
```diff
@@ -177,7 +208,7 @@ script:
     ./configure CXXFLAGS="-W -Wall -Wextra -Werror -Wno-unused-parameter -Wno-maybe-uninitialized -Wno-unknown-warning-option"
     make
   elif [[ "$TRAVIS_BUILD_SYSTEM" == "Makefile" ]]; then
-    make -f Makefile.simple
+    $mingw64 make -f Makefile.simple
```
rpljwi, line 211: I had to hack this line a bit, but I think it's fine, since for non-Windows builds `$mingw64` is just empty.

Quuxplusone: Here, does `make` get expanded out via `export MAKE=mingw32-make` from line 159? I'm guessing not?
I don't like relying on unset env vars to be empty; could this just be conditional on Windows, i.e.,

    elif [[ "$TRAVIS_BUILD_SYSTEM" == "Makefile" && "$TRAVIS_OS_NAME" == "windows" ]]; then
        $MINGW64 make -f Makefile.simple
    elif [[ "$TRAVIS_BUILD_SYSTEM" == "Makefile" ]]; then
        make -f Makefile.simple

.travis.yml
```diff
@@ -164,6 +193,8 @@ before_install:
       echo 'autotools is installed by default'
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew install automake
+    elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
+      $msys2 pacman --sync --noconfirm --needed automake-wrapper
```
Quuxplusone, line 197: I think you should move `export MAKE=mingw32-make` down to here, if the comment is correct that it's only ever used by autotools and not by any of the other build systems.

.travis.yml
```diff
@@ -132,6 +141,22 @@ before_install:
...
+    choco uninstall -y mingw
+    choco upgrade --no-progress -y msys2
+    export msys2='cmd //C RefreshEnv.cmd '
+    export msys2+='& set MSYS=winsymlinks:nativestrict '
+    export msys2+='& C:\\tools\\msys64\\msys2_shell.cmd -defterm -no-start'
+    export mingw64="$msys2 -mingw64 -full-path -here -c "\"\$@"\" --"
+    export msys2+=" -msys2 -c "\"\$@"\" --"
+    $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-toolchain
+    ## Install more MSYS2 packages from https://packages.msys2.org/base here
+    $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-SDL mingw-w64-x86_64-SDL_gfx mingw-w64-x86_64-SDL_mixer mingw-w64-x86_64-SDL_ttf
```
Quuxplusone, line 155: Can you uppercase `$MSYS2` and `$MINGW64` for consistency with all the other env vars?

.travis.yml
```diff
@@ -132,6 +141,22 @@ before_install:
...
+    taskkill //IM gpg-agent.exe //F  # https://travis-ci.community/t/4967
+    taskkill //IM dirmngr.exe //F # this process also seemed to hang after install
+    export PATH=/C/tools/msys64/mingw64/bin:$PATH
```
Quuxplusone, line 158: Could you add a comment here, or even better just explain in your commit message, what this line is for? I.e., what executable is found in `/C/tools/msys64/mingw64/bin` that wouldn't be found without this line? I'm guessing either gcc or make or both, but I'm not sure.

.travis.yml
rpljwi: Ah, you're right. Come to think of it, I didn't use `mingw32-make` locally at all, even with `autotools`. That also reminds me that usual `make` needs to be installed.

.travis.yml
rpljwi: Yes, everything related to the native-build toolchain is there AFAIK. Though I'm not sure that the needed path isn't in the `PATH` already by this point.

rpljwi:
I think that covers everything you've mentioned, except

https://github.com/rpljwi/hyperrogue/blob/6ebd2fa9e752072136a3a92bc7261d2af6325df9/.travis.yml#L158

I'm not sure of the necessity of that, TBH. It's the toolchain binary location, but I don't really know whether it's not exported already by that point. Locally, I didn't use the exact same approach, since there are special binary launchers for Windows that take care of exactly things like that I guess.

I can try looking through docs to clarify.

Quuxplusone:
It seems like Travis isn't happy yet: https://travis-ci.org/github/zenorogue/hyperrogue/jobs/689351528
```
$ # Install GLEW if asked for
  if [[ "$HYPERROGUE_USE_GLEW" == "1" ]]; then
    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
      sudo apt-get install -qq libglew-dev
    elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
      brew install glew
    elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
      $MSYS2 pacman --sync --noconfirm --needed mingw-w64-x86_64-glew
    else
      exit 'Unsupported OS'
    fi
  fi
Refreshing environment variables from registry for cmd.exe. Please wait...Finished..
error: could not open file /var/cache/pacman/pkg/mingw-w64-x86_64-gcc-libs-10.1.0-2-any.pkg.tar.zst: Child process exited with status 127
error: could not open file /var/cache/pacman/pkg/mingw-w64-x86_64-python-3.8.3-1-any.pkg.tar.zst: Child process exited with status 127
error: could not open file /var/cache/pacman/pkg/mingw-w64-x86_64-harfbuzz-2.6.6-1-any.pkg.tar.zst: Child process exited with status 127
error: could not open file /var/cache/pacman/pkg/mingw-w64-x86_64-freetype-2.10.2-2-any.pkg.tar.zst: Child process exited with status 127
error: failed to commit transaction (cannot open package file)
ERROR: The process "gpg-agent.exe" not found.
ERROR: The process "dirmngr.exe" not found.
resolving dependencies...
looking for conflicting packages...
Packages (1) mingw-w64-x86_64-glew-2.2.0-2
Total Download Size:   0.50 MiB
Total Installed Size:  5.44 MiB
:: Proceed with installation? [Y/n] 
:: Retrieving packages...
downloading mingw-w64-x86_64-glew-2.2.0-2-any.pkg.tar.zst...
checking keyring...
checking package integrity...
loading package files...
Errors occurred, no packages were upgraded.
```

The good news is that this seems to be exactly the error message given at https://github.com/msys2/MSYS2-packages/issues/1962 (reported 2 days ago), and so maybe if we just wait a few days everything will be back to normal and Travis will become happy?

I notice the lines
```
ERROR: The process "gpg-agent.exe" not found.
ERROR: The process "dirmngr.exe" not found.
```
in the middle there. I vaguely remember seeing the same help doc that you must have seen, re that first `taskkill` line at least, and deciding that it didn't seem to be needed on Travis. Did you observe that it _was_ needed?

rpljwi:
Added a commit fixing a quirk in `SDL_gfxPrimitives.h`. Pretty much the same fix was required on Arch Linux when I tried (successfully eventually) to build Hyperrogue there. Maybe there is a more "right" way, though.

Re `taskkill`: when installing on Windows through the standalone installer, as per [this instruction](https://www.msys2.org/wiki/MSYS2-installation/) (namely, section V, point 5), I did indeed experience those two programs hanging after an initial core bootstrap/upgrade. It could very well not apply to Chocolatey package -- I should probably investigate.

As for the former errors, now that I look at it more closely, it seems odd that there is no `-y` argument to `pacman`, or no initial `pacman -Syu` at all. To explain a bit (from what you stated elsewhere I assume MacOS is your only system; sorry if I'm explaining the obvious): `-S` a.k.a. `--sync` is the command code for all the install or upgrade operations, `-y` a.k.a. `--refresh` is a subcode for updating package database, and `-u` or `--sysupgrade` is a subcode for, well, upgrading existing packages. It's also not completely clear to me why they went for verbose versions of these commands, since in everyday use you just `pacman -Syu` or something like that.

rpljwi:
Let's see what it thinks about this one. Also, it's a bit unexpected to me that Travis automatically pulls updates to its config even from unapproved changes.

rpljwi:
*sound of forehead slapping*
I forgot that there's noone to press the `Y` button out there. Oops.

rpljwi:
Yeah, I think this was in the wrong place. I can try just one more (admittedly hacky thing), and that's to bypass Chocolatey-package [initial configuration](https://chocolatey.org/packages/msys2), upgrade pacman, and then resume as usual. This might be necessary, because last update to the package was in October of the last year.

rpljwi:
Seems like I was overly optimistic with removing the "draft" status.

For reference, [this here](https://travis-ci.community/t/error-installing-msys2-runtime-zst-package/8687) seems to be the same kind of issue.

Quuxplusone:
In case I was too oblique with my reference to https://github.com/msys2/MSYS2-packages/issues/1962#issuecomment-630258878 , I think the really important comment is this one:

> Sorry for the inconvenience :/ There is some breakage today with the upgrade path from the last stable installer.

In other words, it sounds like the Msys2 maintainers broke something within the past 72 hours, and (they didn't say so but I hope) if we just wait a while they may fix it. I know how tempting it is to keep trying to solve the problem Right Now, but it may be a better use of time to just let this PR sit for a week and then pick it up again and see if it's magically fixed itself.

Your "[this here](https://travis-ci.community/t/error-installing-msys2-runtime-zst-package/8687)" was also posted in the past 24 hours, indicating that this is a very new (and I hope temporary) upstream issue.

rpljwi:
I believe I'd found references to the same problem from earlier than that. But you may be right still.

Sorry for commit-spam, by the way. It's sort of the only way to test, but if the latest one fails, I'll probably stop and wait for the proper fix.

rpljwi:
https://chocolatey.org/packages/msys2#comment-4919416069

OK, I admit that I should've paid more attention. Looks like the problem is indeed very recent, and will probably be fixed soon.

rpljwi:
Looks like (https://github.com/msys2/MSYS2-packages/issues/1970#issuecomment-632716190) the issue has been fixed. I should probably clean up the commits here, and try again.

rpljwi:
It [builds alright](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/690201971) now, but the test fails due to "missing" DLL. Must be something funny going on with the `PATH`. Adding to the earlier question (https://github.com/zenorogue/hyperrogue/pull/116#discussion_r428178794): not only is the toolchain located at `/mingw64/bin`, but all related DLL should be there as well. That is to say, `libstdc++-6.dll` must be there, as long as `mingw-w64-x86_64-gcc-libs` is [installed](https://packages.msys2.org/package/mingw-w64-x86_64-gcc-libs), which, judging by the log, it was.

Quuxplusone:
Could the current failure https://travis-ci.org/github/zenorogue/hyperrogue/jobs/690404626 be due simply to writing `./hyperrogue` instead of `.\hyperrogue.exe`, or some minor thing like that? It's weird that Travis doesn't give any information except "exited with 127."

still-flow:
Actually, it does say a bit more, it's just that for inexplicable reasons that output is disconnected from where it logically belongs: [this line specifically](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/690404626#L282) (it's what I referred to in my last comment).
> C:/Users/travis/build/zenorogue/hyperrogue/hyperrogue.exe: error while loading shared libraries: libstdc++-6.dll: cannot open shared object file: No such file or directory

still-flow:
At last, [it works](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/690859756). Sure took some ugly hacks to pull off though. Now to think up a way to clean up this mess.

.travis.yml
```diff
@@ -203,7 +203,7 @@ before_install:
   fi
 - |-
   # Install autotools if asked for
-  if [[ "$TRAVIS_USE_AUTOTOOLS" == "1" ]]; then
+  if [[ "$TRAVIS_BUILD_SYSTEM" == "autotools" ]]; then
```
still-flow, line 208: Funny. Fixing this [has led](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/691015281) MacOS-autotools build failing because these are apparently pre-installed.

still-flow:
The newly added MinGW-autotools build [fails](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/691038159#L295), but if I read this correctly, it's because it's unique in the fact that it also includes `rogueviz`, while other builds with the same `CXXFLAGS` don't (`-Werror`, chiefly).

On an unrelated note, I don't really like how on build and test steps there have to be checks whether build-host is Windows. Would it be better if, at early preparation steps, `$MINGW64` were just set to an empty string for Unices, so that all the commands could be uniformized?

PR #117, opened 2020-05-21
Proposal to add Github CI

still-flow:
As it turns out, Github has its own CI system as part of "Actions". I'll leave it up to zenorogue to decide whether it's needed or worth it, but here is a more or less direct port of the current Travis configuration. It's not quite the same, of course: the main build matrix is every configuration of
`{MacOS, Ubuntu} x {GCC, clang} x {makefile, autotools, mymake} x {rogueviz, no rogueviz}`,
and `emscripten` build is separate from that matrix. This is just a simple configuration, of course it's up to discussion to decide what needs to be there and what doesn't. There is also a temporary hack disabling warnings->errors. I think that there exist ways to gather all warnings from a build without failing it, but I haven't looked into that yet.

[Example](https://github.com/rpljwi/hyperrogue/actions/runs/111604467) result using the configuration from the commit.

One note to zenorogue: if/when you do decide to add this, consider disabling notifications for actions. Otherwise tests may generate a lot of notifications/emails.

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -1,0 +1,36 @@
+set -e o pipefail
+source .github/workflows/gh_ci_envvars.sh
+
+
+GH_MYMAKE_ARGS=""
+GH_AUTOTOOLS_CXXFLAGS="-W -Wall -Wextra -Wno-unused-parameter -Wno-maybe-uninitialized"
+
+HYPERROGUE_USE_GLEW=$GH_HYP_GLEW
+export HYPERROGUE_USE_GLEW=${HYPERROGUE_USE_GLEW: -1}
+
+HYPERROGUE_USE_PNG=$GH_HYP_PNG
+export HYPERROGUE_USE_PNG=${HYPERROGUE_USE_PNG: -1}
+
+HYPERROGUE_USE_ROGUEVIZ=$GH_HYP_RVIZ
+export HYPERROGUE_USE_ROGUEVIZ=${HYPERROGUE_USE_ROGUEVIZ: -1}
+if [[ "$GH_HYP_RVIZ" == "rviz_1" ]]; then
+  GH_MYMAKE_ARGS+=" -rv"
+  GH_AUTOTOOLS_CXXFLAGS+=" -DCAP_ROGUEVIZ=1"
+fi
+
+export CC=$GH_COMPILER
+export CXX=${CC%cc}++
+
+if [[ "$GH_BUILDSYS" == "makefile" ]]; then
+  make -f Makefile.simple
+elif [[ "$GH_BUILDSYS" == "autotools" ]]; then
+  autoreconf -vfi
+  ./configure CXXFLAGS="${GH_AUTOTOOLS_CXXFLAGS}"
+  make
+elif [[ "$GH_BUILDSYS" == "mymake" ]]; then
+  make -f Makefile.simple mymake
+  ./mymake $GH_MYMAKE_ARGS
+  mv hyper hyperrogue
+else
+  exit 'unknown build system'
+fi
--- a/.github/workflows/github_ci.yml
+++ b/.github/workflows/github_ci.yml
@@ -1,0 +1,80 @@
+name: Github CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  main_matrix:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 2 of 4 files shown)


Quuxplusone:
That's pretty cool! Would there be an equivalent of Travis's red/green badges in the README or otherwise visible?

rpljwi:
> Would there be an equivalent of Travis's red/green badges in the README or otherwise visible?

Yup, there is a "Create status badge" button on the right and a bit below the top in my example which probably does just that.

Github CI
Yep, seems to work. There are various options there which I don't know anything about, but seems snazzy.

rpljwi:
Managed to set up a simple [Windows build](https://github.com/rpljwi/hyperrogue/runs/702342536). Unfortunately, Github's Windows environments are running PowerShell as their main shell, so Windows builds will have to be separate from the Unix matrix.

still-flow:
Hm, I really screwed this PR up, it would seem. Just tried to rebase the proposed changes (and new stuff too) onto the master branch which has since been updated, but this looks like it failed miserably.

still-flow:
And here are [the results](https://github.com/still-flow/hyperrogue/actions/runs/120090616) for the current master. A lot more green, despite disabling the "remove -Werror" crutch. At this point I feel it's ready to leave the draft stage. One problem could be that currently there is a lot of duplication between `build` stages in `windows` and `main_matrix`, which could probably be solved via off-loading these sections in a script, but then there would be a problem of correctly passing envvars into script on Windows.

still-flow:
I couldn't resist the temptation to make things just a bit less inelegant, so here is the result. At least [nothing broke](https://github.com/still-flow/hyperrogue/actions/runs/120202346) compared to the [previous time](https://github.com/still-flow/hyperrogue/actions/runs/120090616). Ubuntu-clang-mymake builds seem to have a [problem](https://github.com/still-flow/hyperrogue/runs/723889068#step:5:12) with the ordering of the log's output, which is somewhat troubling, but that [was present](https://github.com/still-flow/hyperrogue/runs/723601210#step:4:43) without separation into scripts too. I'm not sure what to do with that.

still-flow:
@zenorogue, would you be interested in this? AFAIK this works as it should now, here is an [example of the output](https://github.com/still-flow/hyperrogue/actions/runs/120202346) (mind you, my `master` is pretty old right now, so there are more errors than there would be now).

zenorogue:
Thanks! I think it looks nicer than the current Travis configuration. Would that replace Travis, or is there a point to keep both? I assume the badge needs to be replaced/added to README.md, is there anything else to do?

still-flow:
The only thing that's sort of different from Travis is that I wasn't able to find a way to display different badges for different build targets (e.g. "web", "autotools", etc.). I honestly don't know if one is better than the other, however Github seems to be a bit faster? And maybe "the more, the better" philosophy applies in this case, too.

zenorogue:
It seems that some targets currently fail -- as far as I can tell, this is because RogueViz requires c++17. With mymake it should be sufficient to add an option `-std=c++17`, not sure about autotools.

Quuxplusone:
FWIW, I'd agree that "the more, the better."  (And I continue to think that if autotools is giving you trouble, just throw it overboard. ;))

One downside of GitHub CI seems to be that it very quickly recycles the logs: [this](https://github.com/still-flow/hyperrogue/actions/runs/120202346) says "The logs for this run have expired and are no longer available." Whereas in TravisCI you can currently go back [six months](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/668379008) or more.

Also, Travis's "build history" seems quite discoverable and intuitive, whereas I can't figure out how to navigate or really do anything in GitHub CI. But that might just be a side-effect of having the CI stuff in a branch instead of master, I dunno.

still-flow:
> With mymake it should be sufficient to add an option `-std=c++17`

Would you like me to add it to workflow config?

> One downside of GitHub CI seems to be that it very quickly recycles the logs

I was able to find [this topic](https://github.community/t/no-log-expiration-setting-for-github-actions/127787) from a month ago in which this very issue is raised. The reply suggests giving feedback through a special form, which OP presumably used, but maybe it's worth to use it again?

On a related note, it's possible to [download logs manually](https://docs.github.com/en/actions/managing-workflow-runs/using-workflow-run-logs#downloading-logs) (which btw apparently contain more information than is available via GUI), and a shoddy workaround could be to somehow save that data as a build artifact. The trouble is that there is no obvious way to get the download link from *within* the build environment.

zenorogue:
> Would you like me to add it to workflow config?

Yes please!

PR #118, opened 2020-05-21
Add basic parallel building to mymake via shell jobs system?

still-flow:
An attempt to speed up `mymake` builds a bit. Done via the possibly stupid (and probably not Windows-compatible) way of using the shell's job system. There was indeed some speedup locally for me, and here is the comparison of Github CI results:
- [serial mymake](https://github.com/rpljwi/hyperrogue/actions/runs/111604467)
- [parallel mymake](https://github.com/rpljwi/hyperrogue/actions/runs/111916309)

As you can see, overall run time went down somewhat. Of course, this will need further testing. Again, it's not up to me to decide whether it's worth it.

Oh, and there is no idiot check for `-j` parameter -- not sure if it's needed.

### DIFF
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -19,6 +19,8 @@ using namespace std;
 
 string opts;
 
+int batch_size = 6; // how many modules to compile in parallel
+
 string default_standard = " -std=c++11";
 string standard = default_standard;
 
@@ -129,6 +131,8 @@ int main(int argc, char **argv) {
       standard = s;
     else if(s.substr(0, 2) == "-l")
       linker += " " + s;
+    else if(s.substr(0, 2) == "-j")
+      batch_size = stoi(s.substr(2));
     else if(s == "-I") {
       opts += " " + s + " " + argv[i+1];
       i++;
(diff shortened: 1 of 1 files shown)


Quuxplusone:
(Takes the CI build from 16 minutes to 12 minutes. Doesn't feel worth-it to me, FWIW.)
Yet a different approach would be to spin off each `system()` into its own `std::thread`. I wonder whether that would complexify or simplify the code, and how it would affect the running time.

rpljwi:
Another approach could be to emit a Makefile and use `make`'s parallel build capability.

rpljwi:
I think I like the delegation approach more; the code is quite simple. Here is the comparison (reusing entries from OP):
- [serial mymake](https://github.com/rpljwi/hyperrogue/actions/runs/111604467)
- [shell-parallelized mymake](https://github.com/rpljwi/hyperrogue/actions/runs/111916309)
- [make-parallelized mymake](https://github.com/rpljwi/hyperrogue/actions/runs/699029344)

The last two aren't exactly the same, since the very last entry uses `-j4` as opposed to `-j6`, but the time is even a bit better now.

I should probably close this PR and open one with that new approach, right?

rpljwi:
As for threads, well, I have close to none experience programming with them, and I hear that their wrong usage can lead to some subtle bugs, so it's probably best left to someone else.

rpljwi:
One last thing: on my system (4 cores) usual `mymake` takes around 5:50, and the one delegating to `make` (_not_ the one above -- haven't made the same comparison with that) takes only about 3:00. This is after running a couple of times.

still-flow:
Closing this in favor of another approach. Will post a PR once that's in at least somewhat presentable state.

PR #121, opened 2020-05-28
Add parallel compilation to mymake

still-flow:
Dunning-Kruger effect is a really powerful thing, I guess. After discussing threaded approach in #118, I eventually decided that it's not as hard as it seems, and here is what I've come up with. Since I'm more out of my depth than usual here, all suggestions and corrections are especially welcome.

Before anything else, of course I tested it a couple of times, and nothing seemed to break thus far (including correct resumption after `SIGINT`). Total compilation time (including Rogueviz) on my 4-core system went down from 7:50 to 3:50 if using `-j` option, which should be equivalent to batch size of `$(nproc)+1`, which is often suggested for make.

### DIFF
--- a/Makefile.simple
+++ b/Makefile.simple
@@ -162,7 +162,7 @@ savepng$(OBJ_EXTENSION): savepng.cpp
 	$(CXX) -O2 $(CXXFLAGS) -c savepng.cpp -o $@
 
 mymake$(EXE_EXTENSION): mymake.cpp
-	$(CXX) -O2 $(CXXFLAGS) mymake.cpp -o $@
+	$(CXX) -O2 $(CXXFLAGS) -lpthread mymake.cpp -o $@
 
 emscripten: hyper.html
 
--- a/Makefile.simple
+++ b/Makefile.simple
@@ -162,7 +162,7 @@ savepng$(OBJ_EXTENSION): savepng.cpp
 	$(CXX) -O2 $(CXXFLAGS) -c savepng.cpp -o $@
 
 mymake$(EXE_EXTENSION): mymake.cpp
-	$(CXX) -O2 $(CXXFLAGS) -lpthread mymake.cpp -o $@
+	$(CXX) -O2 $(CXXFLAGS) mymake.cpp -lpthread -o $@
 
 emscripten: hyper.html
 
--- a/Makefile.simple
+++ b/Makefile.simple
@@ -162,7 +162,7 @@ savepng$(OBJ_EXTENSION): savepng.cpp
 	$(CXX) -O2 $(CXXFLAGS) -c savepng.cpp -o $@
 
 mymake$(EXE_EXTENSION): mymake.cpp
-	$(CXX) -O2 $(CXXFLAGS) mymake.cpp -lpthread -o $@
+	$(CXX) -O2 $(CXXFLAGS) mymake.cpp -pthread -o $@
 
 emscripten: hyper.html
 
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -210,6 +210,8 @@ int main(int argc, char **argv) {
   
   string allobj = " " + obj_dir + "/hyper.o";
 
+  printf("compiling modules using batch size of %d:\n", batch_size);
+
   int id = 0;
   vector<pair<int, string>> tasks;
   for(string m: modules) {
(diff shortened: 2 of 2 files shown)


Makefile.simple
```diff
@@ -162,7 +162,7 @@ savepng$(OBJ_EXTENSION): savepng.cpp
 	$(CXX) -O2 $(CXXFLAGS) -c savepng.cpp -o $@
 
 mymake$(EXE_EXTENSION): mymake.cpp
-	$(CXX) -O2 $(CXXFLAGS) mymake.cpp -o $@
+	$(CXX) -O2 $(CXXFLAGS) mymake.cpp -lpthread -o $@
```
Quuxplusone, line 165: Nit: You should use the compiler option `-pthread` instead of `-lpthread` here; `-pthread` adds some extra macro definitions such as `_REENTRANT` which _might_ matter to some library code.

mymake.cpp
```diff
@@ -218,14 +226,45 @@ int main(int argc, char **argv) {
...
+        ++tasks_done;
+        goto check_lollygagging;
+        }
+      }
+    else if (tasks_taken < tasks_amt) {
+      auto task = tasks[tasks_taken];
+      int mid = task.first;
+      string cmdline = task.second;
+      printf("compiling %s... [%d/%d]\n", modules[mid].c_str(), tasks_taken+1, tasks_amt);
+      worker = async(launch::async, (int (*)(string))system, cmdline);
```
Quuxplusone, line 264: I recommend making `tasks` a vector of `std::function<void(int, int)>`; then replace these 5 lines with
```
worker = std::async(std::launch::async, tasks[tasks_taken], tasks_taken+1, tasks_amt);
```
and replace lines 229–230 with
```
std::string cmdline = compiler + " " + opts + " " + src + " -o " + obj;
tasks.push_back([m, cmdline](int x, int y) {
    printf("compiling %s... [%d/%d]\n", m.c_str(), x, y);
    system(cmdline);
});
```
This eliminates the UB in `(int (*)(string))system` (namely, you're not allowed to assume that the standard library function has an overload with that exact signature).

Packaging up the `system` call into a `std::function` is the main thing. I admit it might be cleaner to keep the module name `m` separately outside the function (i.e. `std::pair<std::string, std::function<void()>>`), rather than passing `tasks_taken` and `tasks_amt` back _into_ the function.

Makefile.simple
still-flow: Sure thing. One question, though: it's `-lpthread` everywhere else (in `Makefile.simple`, as well as in `opts` in `mymake.cpp`) -- should it be fixed there as well, or is there a specific reason for that?

Quuxplusone: Oh yeah, I would indeed advise changing it to `-pthread` everywhere (and moving it into `CXXFLAGS_EARLY` in Makefile.simple, so that it's seen by the compiler as well as the linker). That might warrant a separate patch.

mymake.cpp
still-flow: Okay, that conversion did indeed feel quite ugly to me, however in this case `system(string)` is a locally-defined function, so (correct me if that's wrong) I guess `system` can be assumed to have this overload?
https://github.com/zenorogue/hyperrogue/blob/4386f86efb88c065fd82b3bf42fc8ba8f88d3eb0/mymake.cpp#L73-L75

I'll probably need some more time to process this suggestion. It does seem more elegant, though it needs at least a minor change because the return value of `system` call is of interest. And I don't imagine that can be packaged into a lambda (as a simple `printf` message, at least) because right now the behavior is that once one module fails compilation, everything stops; it feels like a good idea to keep this behavior.

Actually, if we are to go the route of packing the compilation-notification message into a lambda, why not just pre-bake the string-to-be-`printf`d?

Makefile.simple
still-flow: Looked up GCC's man-page, and indeed it states re `-pthread`:
> You should use this option consistently for both compilation and linking.

mymake.cpp
still-flow: I thought about this for a bit and decided against packing messages into `std::function`. So that a `task` is a `std::pair<int, std::function<int(void)>>`. I'll commit the changes shortly.

still-flow:
Hey, @zenorogue, may you be interested in this? I think the most obvious problems have been fixed by now. Perhaps I should squash the commits beforehand?

zenorogue:
Yeah, I assumed that this was still work in progress. Very cool indeed. Thanks!

PR #122, opened 2020-06-02
Eliminate the option to pass -DCAP_PNG=0 on desktop.

Quuxplusone:
CAP_TEXTURE now relies on CAP_PNG, and the whole thing relies on CAP_TEXTURE.
Passing `-DCAP_PNG=0` just makes the build fail.

Being tired of having my local build (on Mac OSX) fail for lack of `CAP_PNG`, I'm using this patch locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ matrix:
       TRAVIS_COMPILER_NAME=gcc
       TRAVIS_BUILD_SYSTEM=autotools
       HYPERROGUE_USE_GLEW=1
-      HYPERROGUE_USE_PNG=1
   - os: osx
     compiler: clang
     env: >-
@@ -18,7 +17,6 @@ matrix:
       TRAVIS_COMPILER_NAME=clang
       TRAVIS_BUILD_SYSTEM=autotools
       HYPERROGUE_USE_GLEW=1
-      HYPERROGUE_USE_PNG=1
   - os: linux
     compiler: gcc
     env: >-
@@ -26,7 +24,6 @@ matrix:
       TRAVIS_COMPILER_NAME=gcc
       TRAVIS_BUILD_SYSTEM=Makefile
       HYPERROGUE_USE_GLEW=1
-      HYPERROGUE_USE_PNG=1
   - os: linux
     compiler: clang
     env: >-
@@ -34,7 +31,6 @@ matrix:
       TRAVIS_COMPILER_NAME=clang
       TRAVIS_BUILD_SYSTEM=Makefile
       HYPERROGUE_USE_GLEW=1
-      HYPERROGUE_USE_PNG=1
   - os: linux
     dist: bionic
     compiler: gcc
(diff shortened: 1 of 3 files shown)


Quuxplusone:
@zenorogue ping? The Travis build is green (hooray!) because Travis has not tested the `CAP_PNG=0` codepath since a708c10f1d9a5fb9b9cd93b4ff360cd9449bb987. But, locally, when I do `make -f Makefile.simple` on Mac OSX, it still defaults to `CAP_PNG=0` and breaks my local build.

So I'd like to get this patch committed. (Or, if you are working on re-enabling `CAP_PNG=0` mode, that would be good to know; but I doubt you are _and_ I doubt it'd be worth the trouble.)

Quuxplusone:
Btw, I would be interested in submitting a patch to remove `CAP_TEXTURE` (that is, to hardcode it to 1 and constant-propagate that 1 to eliminate all the `#if`s that depend on it), and/or to remove `ISMINI` (that is, to hardcode it to 0 and constant-propagate). I don't think the current codebase can possibly support any value for `ISMINI` except 0, or any value for `CAP_TEXTURE` except 1. Am I wrong?

zenorogue:
While I do not check these flags regularly to make sure that all combinations work, I would not like to remove them -- for example, I recently wanted to run some computational experiments using the HyperRogue engine on a computer that has lots of RAM but no SDL or OpenGL, and "-DCAP_GL=0 -DCAP_SOLV=0 -DCAP_SDL=0 -DCAP_PNG=0 -DCAP_TIMEOFDAY=1" was useful for this. (With a small problem that mymake still requests to link all the unused libraries...)

It currently does compile correctly with -DCAP_PNG=0 for me -- probably because I have fixed some guards to make the above compile.

Also I think Makefile.simple should default to CAP_PNG=1.

Quuxplusone:
> I recently wanted to run some computational experiments using the HyperRogue engine on a computer that has lots of RAM but no SDL or OpenGL, and "-DCAP_GL=0 -DCAP_SOLV=0 -DCAP_SDL=0 -DCAP_PNG=0 -DCAP_TIMEOFDAY=1" was useful for this.

I think your recent "fixup guards" commits helped a lot. There are still some combinations that don't work, e.g. `CAP_GL` without `CAP_SDL`, or `CAP_SDL` without `CAP_GL`... but it looks like `CAP_PNG` is now fixed.

Passing `-DCAP_SHOT=0` still makes everything break. (Equivalently, passing `-DCAP_PNG=0 -DCAP_SVG=0` makes everything break.)

> Also I think Makefile.simple should default to CAP_PNG=1.

That's what the current PR does; but now that you've fixed the guards so that `-DCAP_PNG=0` works again, I think it would make more sense to just add a new entry to the TravisCI build matrix verifying that `-DCAP_PNG=0` continues to work in the future. At which point I don't think it really matters what the default is in Makefile.simple.

Quuxplusone:
Abandoning as obsolete/resolved.

PR #127, opened 2020-09-22
Fix some Github CI red checks

still-flow:
([related](https://github.com/zenorogue/hyperrogue/pull/117#issuecomment-696108807) to #117)


still-flow:
Besides borked Windows packaging stuff, the only issues remaining are [with clang builds](https://github.com/zenorogue/hyperrogue/pull/127/checks?check_run_id=1150695548#step:5:137). Thankfully, the error message is rather helpful and suggests a fix: add `-fPIC` to compiler options.

still-flow:
Now, Windows stuff can be easily (I believe) fixed by adding a step to upgrade packages before installing dependencies, however as a downside Windows builds would become somewhat slower. According to [this](https://github.com/marketplace/actions/setup-msys2), Windows images are updated approximately every 10 days, so this is the tradeoff here. Is it worth to add a bit of overhead per build so that you don't have to potentially wait 10 days until the image gets updated? I'm personally leaning towards "yes".

still-flow:
This didn't work as intended. However, the better solution probably is to use the Github Action mentioned above. This will require rewriting the workflow config somewhat.

still-flow:
Alright, now only two Windows builds fail, both for the same reason: [some missing declarations](https://github.com/zenorogue/hyperrogue/pull/127/checks?check_run_id=1151665824#step:6:14) for `fork`, `wait` and `pipe`. Seems bizarre that it only fails on Windows, but then again these three are Unix stuff AFAIK?

zenorogue:
Indeed this part was supposed to only work in Linux (for now) -- I have added CAP_VIDEO guards, which should fix the problem. Thanks!

PR #128, opened 2020-09-23
Improve CI configuration in various ways

still-flow:
Primarily addresses #115 (adding Windows-mymake build matrix entry and catching gdb backtraces for it), but also includes other sundry improvements; see commit message for more details.


still-flow:
Only two builds fail, which is to be expected: the aforementioned PR isn't yet merged into `master` so `mymake` isn't supposed to work on Windows at all.

PR #129, opened 2020-09-25
Remove USE_UNORDERED_MAP because it has bit-rotted.

Quuxplusone:
Trying to compile with `-DUSE_UNORDERED_MAP` produces lots of compiler errors
like these, because of missing `std::hash` specializations.
Also, `#define unordered_map map` is just evil!

```
./nonisotropic.cpp:875:36: note: in instantiation of template class 'std::__1::unordered_map<hr::nilv::mvec, hr::heptagon *,
      std::__1::hash<hr::nilv::mvec>, std::__1::equal_to<hr::nilv::mvec>, std::__1::allocator<std::__1::pair<const hr::nilv::mvec, hr::heptagon
      *> > >' requested here
    unordered_map<mvec, heptagon*> at;
                                   ^

./nonisotropic.cpp:239:58: note: in instantiation of template class 'std::__1::unordered_map<std::__1::pair<hr::heptagon *, hr::heptagon *>,
      hr::heptagon *, std::__1::hash<std::__1::pair<hr::heptagon *, hr::heptagon *> >, std::__1::equal_to<std::__1::pair<hr::heptagon *,
      hr::heptagon *> >, std::__1::allocator<std::__1::pair<const std::__1::pair<hr::heptagon *, hr::heptagon *>, hr::heptagon *> > >'
      requested here
    unordered_map<pair<heptagon*, heptagon*>, heptagon*> at;
                                                         ^

./nonisotropic.cpp:457:49: error: no matching member function for call to 'iadj'
      while(h1->distance < h2->distance) back = iadj(h2, down) * back, h2 = h2->cmove(down);
                                                ^~~~
cell.cpp:42:15: note: candidate function not viable: no known conversion from 'hr::sn::hrmap_solnih' to 'hr::hrmap' for object argument
  transmatrix iadj(heptagon *h, int d) {
              ^
cell.cpp:41:22: note: candidate function not viable: no known conversion from 'hr::sn::hrmap_solnih' to 'hr::hrmap' for object argument
  struct transmatrix iadj(cell *c, int i) { cell *c1 = c->cmove(i); return adj(c1, c->c.spin(i)); }
                     ^
```

### DIFF
--- a/asonov.cpp
+++ b/asonov.cpp
@@ -136,8 +136,8 @@ EX transmatrix adjmatrix(int i) {
   }
   
 struct hrmap_asonov : hrmap {
-  unordered_map<coord, heptagon*> at;
-    unordered_map<heptagon*, coord> coords;
+  map<coord, heptagon*> at;
+    map<heptagon*, coord> coords;
     
   heptagon *getOrigin() override { return get_at(coord(0,0,0)); }
     
--- a/basegraph.cpp
+++ b/basegraph.cpp
@@ -22,7 +22,7 @@ struct display_data {
   /** The view relative to the player character. */
   shiftmatrix player_matrix;
   /** On-screen coordinates for all the visible cells. */
-  unordered_map<cell*, shiftmatrix> cellmatrices, old_cellmatrices;
+  map<cell*, shiftmatrix> cellmatrices, old_cellmatrices;
   /** Position of the current map view, relative to the screen (0 to 1). */
   ld xmin, ymin, xmax, ymax;
   /** Position of the current map view, in pixels. */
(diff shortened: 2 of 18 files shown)


Quuxplusone:
@zenorogue ping!

Quuxplusone:
@zenorogue ping!

zenorogue:
More like an unfinished feature... I think std::unordered_map should be more efficient than std::map, so the correct solution here IMO would be to add the missing specializations, and either decide that one of the options is useless and remove it (which probably would be std::map), or introduce a new type name which is equal to std::map or std::unordered_map depending on this option. But yes, probably good to merge this for now.

Quuxplusone:
Thanks!
IMO, if you were going to pursue this again, the right approach would be like what I did with `hookset` in d1458b3ee5988d16dfb4920e81d52b216c42daee. Exposing a _typedef_ for either `std::map` or `std::unordered_map` is still too weakly typed; you should wrap the `std::map` up as a private data member of some class type which exposes a closed set of public functionality, so that there's only O(1) amount of code to modify in order to switch from `map` to `unordered_map`.
That doesn't solve _all_ your problems (like, should you enforce that the key type must be hashable, while still using `map`? should you enforce that it's `<`-able, afterward?) but it'd be a lot safer path forward.

PR #130, opened 2020-09-25
Add "hyper.html" to .gitignore

Quuxplusone:
`hyper.html` is created as part of the build process, and therefore shouldn't be checked in; but it seems that it was checked in anyway, in 1ac2805b7f38784cd61566d137e8e3b2a30b8a09.
If we can't stop it from being checked in, at least we can gitignore it so that it doesn't cause bogus `git diff` output.

This PR is just the silly solution to stop `git diff` from producing hundreds of lines of "hey, `make clean` deleted this file" every time you make clean.

Of course, if we _can_ stop `hyper.html` from being checked in, let's do that!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ config.log
 config.status
 configure
 depcomp
+hyper.html
 Makefile
 install-sh
 missing


still-flow:
> if we _can_ stop `hyper.html` from being checked in

Unless I'm missing something, wouldn't simple `git rm` be just fine? Sure, it'll remain in older commits, but what of it?

Quuxplusone:
> Unless I'm missing something, wouldn't simple `git rm` be just fine?

In an ideal world, yes, definitely. But `hyper.html` was already not-in-the-repo, prior to 1ac2805, so the mere fact of being not-in-the-repo didn't stop it from getting added to a commit by accident. (At least I assume it was by accident. If it was on purpose, then maybe `git rm` plus a reminder not to commit build products _will_ suffice, and I am happy to update the PR. :))

Btw, compare `hyper.exe`, which was part of the initial commit 9b11f1b6c55ce874ca20909ac8343b184b52eb3e. I `git rm`'ed it in 0946300663d1975ffba0935ac91d2b9bd441b7d4, and it hasn't come back yet. So that's good.

Another alternative, using technological measures, would be to `git rm hyper.html` and then _add a CI step that will fail the build_ anytime `make -f Makefile.simple clean` on a fresh checkout produces any diffs.

still-flow:
> mere fact of being not-in-the-repo didn't stop it from getting added to a commit by accident

Hm... So I see at least two ways out of this for the future:
- `.gitignore` all possible build artifacts. Easier to do right now, but will need manual update in case build artifacts change.
- Out-of-tree builds. Will require considerably more rewriting, but more future-proof I imagine?

> _add a CI step that will fail the build_ anytime `make -f Makefile.simple clean` on a fresh checkout produces any diffs

This feels just wrong, somehow. Probably because it would require rewriting stuff in more than one place if anything changes, which is easy to neglect.

Also, please see my updated reply above. Sorry, "good idea occurs only after the fact", as the proverb goes.

Quuxplusone:
>> add a CI step that will fail the build anytime `make -f Makefile.simple clean` on a fresh checkout produces any diffs

> This feels just wrong, somehow. Probably because it would require rewriting stuff in more than one place if anything changes, which is easy to neglect.

I'm not sure I follow. If we add a new build artifact, we have to change `Makefile.simple` in two places: once to say how the artifact is produced, and once to add it to the `clean` rule. (This is already the case today.)

The only time we'd have to touch the new CI step would be if we switch off of `Makefile.simple` entirely; like if the blessed "remove artifacts" step became `mymake clean` then we'd have to switch the CI step to use `mymake clean` instead of `make -f Makefile.simple clean`.

> [Alternatively] `.gitignore` all possible build artifacts

This _would_ require updating two different places every time we add a build artifact: we'd have to update both `Makefile.simple` and `.gitignore`.

> [Alternatively] Out-of-tree builds

We already kinda have that with `mymake`'s `mymake_files/` directory, which is isomorphic to CMake's idiomatic `build` directory. This has the same technological problem, in principle: what happens if someone accidentally does a `git add mymake_files/` or `git add build/`?

> FWIW I prefer to use `git clean -df[x]` instead of `make clean` lately.

That works if you have good hygiene, i.e. in the ideal world that we are clearly not in, in this specific case. :)  You're basically saying "I never commit build artifacts, _therefore_, when I want to delete all build artifacts, I can just `git clean`." That's actually a good life hack. But the problem is that the statement is tautological: If you ever accidentally _do_ commit a build artifact, then `git clean` will _not_ remove it — effectively, it's a seatbelt that works _right up until_ the car crash, at which point by definition it stops working. If you want to enforce that build artifacts are never committed, you have to somehow tell the computer about the list of build artifacts, so that it has something to enforce. Right now, the list of build artifacts is encoded in `Makefile.simple`'s `clean` production (and nowhere else).

I'm going to take an hour sometime today and redo this PR to be my "CI step" thing, so that we have a concrete proposal to discuss. Stay tuned. :)

PR #69, opened 2019-01-07
Add -Wimplicit-fallthrough

Quuxplusone:
Clang doesn't enable this warning by default, so we pass it explicitly.
GCC pre-GCC 7 doesn't recognize the flag, so we can't pass it for GCC.
GCC 7 enables it by default as part of -Wextra.

Fixes #68.

PR #131, opened 2020-09-29
New round of CI improvements/fixes

still-flow:
Some small fixups:
- exit->quit in gdb cmdlist (funny that it sorta worked regardless)
- remove unnecessary backtrace gdb cmd (the file and line at fault are already reported without it)
- remove mymake output buffering for all platforms (should help with on-the-fly CI logs)
- remove `brew update` step (to [address](https://github.com/zenorogue/hyperrogue/pull/69#issuecomment-700277990) #69)
- move from [to-be-deprecated](https://hub.docker.com/r/trzeci/emscripten/) emscripten container

Now some stuff in need of more explanation.

Switch to `podman` instead of `docker`: this one is up to debate, but [several](https://medium.com/technopanti/docker-vs-podman-c03359fabf77) [articles](https://xperimentalhamid.com/devops/podman-vs-docker/) state that one is "better" than the other. I'll be honest and say that I'm not really sure whether all that benefits this project at all, but then again we only use `docker` in the simplest possible way.

Add a helper log-saver workflow: [addressing](https://github.com/zenorogue/hyperrogue/pull/117#issuecomment-696123813) the issue raised by @Quuxplusone in #117, this should be a temporary workaround until Github perhaps introduces a better way of doing this. For now, I had to create a helper workflow that's triggered after the main one completes, and fetches its logs using [Github's API](https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#download-workflow-run-logs), and uploads them as an artifact. It also displays a link back to the main run in the "annotations" section -- it'd be better if the main run could link to the helper workflow run instead, but there's a chicken-and-egg problem: you can only use the API query after the workflow completes, and by that point AFAIK you can't modify its "executive overview" output. [Example output](https://github.com/still-flow/hyperrogue/actions/runs/273249388). Btw, feel free to scrutinize the hell out of that one -- these rudimentary scripts were hacked together only to the point of working, not to the point of being "correct".

Note that the CI run for this PR won't produce any log archives yet, because [the corresponding `.yml` file has to exist on `master`](https://github.community/t/on-workflow-run-does-not-work-for-me/128833/2).

### DIFF
--- a/.github/workflows/github_ci.yml
+++ b/.github/workflows/github_ci.yml
@@ -77,7 +77,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build
       run: |
-        docker run --rm -v $(pwd):/src trzeci/emscripten make -f Makefile.simple emscripten
+        podman run --rm -v $(pwd):/src emscripten/emsdk make -f Makefile.simple emscripten
     - name: Do a simple test
       run: |
         ls -lAF hyper.html hyper.js hyper.wasm
--- a/.github/workflows/github_ci_logs.yml
+++ b/.github/workflows/github_ci_logs.yml
@@ -1,0 +1,32 @@
+name: Github CI log archiver
+on:
+  workflow_run:
+    workflows: ["Github CI"]
+    types:
+      - completed
+
+jobs:
+  save_logs:
+    name: Save run logs
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download log archive
+      run: |
+        curl -sI -u ${{ github.repository_owner }}:${{ github.token }} \
+          https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.event.workflow_run.id }}/logs > response.txt
+        DOWNLOAD_LINK="https:"$(cat response.txt | grep -Fi Location: | cut -d: -f3)
+        DOWNLOAD_LINK=${DOWNLOAD_LINK%$'\r'}
+        curl -JO "$DOWNLOAD_LINK"
+    - name: Unpack log archive
+      id: archive_unpack
+      run: |
+        LOG_FILENAME=$(ls logs* | cut -d. -f1)
+        echo "::set-output name=log_filename::$LOG_FILENAME"
+        7z x -ologs "$LOG_FILENAME".zip
+    - name: Upload log archive as an artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: ${{ steps.archive_unpack.outputs.log_filename }}
+        path: logs/**
+    - name: Print the link to the base run
+      run: echo "::warning::This log was saved from https://github.com/${{ github.repository }}/actions/runs/${{ github.event.workflow_run.id }}"
--- a/.github/workflows/install_deps.sh
+++ b/.github/workflows/install_deps.sh
@@ -26,7 +26,6 @@ fi
 if [[ "$GH_OS" == "ubuntu-latest" ]]; then
   sudo apt-get -y install $GH_DEPS_UBUNTU
 elif [[ "$GH_OS" == "macos-latest" ]]; then
-  brew update
   # macos-latest already has both gcc and clang
   brew install $GH_DEPS_MACOS
   # work around https://stackoverflow.com/questions/51034399/ for now
--- a/mymake.cpp
+++ b/mymake.cpp
@@ -57,7 +57,6 @@ void set_mingw64() {
   linker = "g++ -o hyper";
   opts = "-DWINDOWS -DCAP_GLEW=1 -DCAP_PNG=1";
   libs = " savepng.o hyper.res -lopengl32 -lSDL -lSDL_gfx -lSDL_mixer -lSDL_ttf -lpthread -lz -lglew32 -lpng";
-  setvbuf(stdout, NULL, _IONBF, 0); // MinGW is quirky with output buffering
   }
 
 vector<string> modules;
@@ -84,6 +83,7 @@ bool file_exists(string fname) {
   }
   
 int main(int argc, char **argv) {
+  setvbuf(stdout, NULL, _IONBF, 0); // this should help with responsiveness of the real-time CI logs
 #if defined(MAC)
   set_mac();
 #elif defined(WINDOWS)
--- a/.github/workflows/test_simple.sh
+++ b/.github/workflows/test_simple.sh
@@ -5,8 +5,7 @@ if [[ "$GH_OS" == "windows-latest" && "$GH_BUILDSYS" == "mymake" ]]; then
 
 cat << ENDOFCMDS > .github/workflows/gdb_cmds.txt
   run --version
-  backtrace
-  exit 1
+  quit 1
 ENDOFCMDS
 
   gdb --batch -x .github/workflows/gdb_cmds.txt ./hyperrogue


.github/workflows/github_ci.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build
       run: |
-        docker run --rm -v $(pwd):/src trzeci/emscripten make -f Makefile.simple emscripten
+        podman run --rm -v $(pwd):/src emscripten/emsdk make -f Makefile.simple emscripten
```
Quuxplusone, line 80: Could you please update `.travis.yml` as well?

.github/workflows/github_ci_logs.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  save_logs:
+    name: Save run logs
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download log archive
+      run: |
+        curl -sI -u ${{ github.repository_owner }}:${{ github.token }} \
+          https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.event.workflow_run.id }}/logs > response.txt
+        DOWNLOAD_LINK="https:"$(cat response.txt | grep -Fi Location: | cut -d: -f3)
```
Quuxplusone, line 17: Bash style nit: I'd prefer `"https:$(cat ...)"` rather than `"https:"$(cat ...)`. I think they end up meaning exactly the same thing, even in the presence of special characters; but the latter feels saner and clearer to me that the _whole thing_ is supposed to be a string.

.github/workflows/github_ci_logs.yml
```diff
@@ -0,0 +1,32 @@
...
+        DOWNLOAD_LINK=${DOWNLOAD_LINK%$'\r'}
```
Quuxplusone, line 18: I had to look up what this line is doing. (`$'\r'` is a literal carriage return, and `${x%y}` trims a suffix, so this line is trimming any trailing carriage-return from the end of the `Location:` header extracted on the line above.)

I wonder if the above line could just be

    DOWNLOAD_LINK="https:$(cat response.txt | grep -Fi Location: | cut -d: -f3 | tr -d $'\r')"

or even

    DOWNLOAD_LINK="$(cat response.txt | grep -Fi Location: | cut -d: -f2 | tr -d $'\r')"

Is part of the intent that you need to change a literal `http` in the response to an `https` in the download link?

.github/workflows/install_deps.sh
```diff
@@ -26,7 +26,6 @@ fi
 if [[ "$GH_OS" == "ubuntu-latest" ]]; then
   sudo apt-get -y install $GH_DEPS_UBUNTU
 elif [[ "$GH_OS" == "macos-latest" ]]; then
-  brew update
   # macos-latest already has both gcc and clang
```
Quuxplusone, line 29: Might want to drop a comment here like

    # As of 2020-09-30, brew update sometimes times out on Github CI, so we don't do it here

and a link to a bug report/SO question, if there is any. (It sounded like you were trying to throw that ball into my court, but so far I've ignored it. ;) I don't really know how to file issues about Github CI; seems like maybe one would just post on https://github.community/search?q=%22brew%20update%22 ?)

mymake.cpp
```diff
@@ -84,6 +83,7 @@ bool file_exists(string fname) {
   }
   
 int main(int argc, char **argv) {
+  setvbuf(stdout, NULL, _IONBF, 0); // this should help with responsiveness of the real-time CI logs
```
Quuxplusone, line 86: As long as you're touching it anyway, how about `s/NULL/nullptr/`? :)

Quuxplusone (commented):
LGTM mod my comments. The only really _really_ important one is the first one, about updating `.travis.yml` with the new Docker image and/or `s/docker/podman/` so that it stays in sync with the Github workflow.

.github/workflows/github_ci.yml
still-flow: Oh, sure, I've been meaning to give Travis some attention after solving the more outstanding issues with the GH CI. Shouldn't be hard to do using the latter as a template.

.github/workflows/github_ci_logs.yml
still-flow: > `$'\r'` is a literal carriage return, and `${x%y}` trims a suffix, so this line is trimming any trailing carriage-return from the end of the `Location:` header extracted on the line above.

Yep, had to add all that because `curl` didn't like the trailing CR (thanks, someone at StackOverflow).

> Is part of the intent that you need to change a literal `http` in the response to an `https` in the download link?

Nah, it's just that `grep -Fi Location:` produces a string that contains _two_ colons: "Location: https://{whatever}", so this is a lazy workaround with `cut`. I suppose something like `awk` would be a better fit for this.

Quuxplusone: Ah, right, you have to change a leading `" https"` to `"https"`. This is the sort of problem I always want to use `grep -o` for, and it usually doesn't quite work, and I end up using `sed` instead. (I don't know `awk`, unfortunately.)

Does this work? (Slightly more hard-coded than yours. Although they'd be within their protocol rights to do so, I really hope Google isn't sending you a `Location:` header spelled like `LoCaTiOn:`. 😜)

    DOWNLOAD_LINK="$(cat response.txt | grep '^Location:' | grep -o 'https://\S*')"

mymake.cpp
still-flow: Aye, I forgot that we're in C++ land here.

.github/workflows/install_deps.sh
still-flow: > It sounded like you were trying to throw that ball into my court, but so far I've ignored it.

Yes and no :). Much of the things I write here are also for myself, because memory/attention are unreliable.
I think https://github.community is just "one user helps another" type of deal. There was a special form for contacting Github support though. I'll see about filing a report later.

.github/workflows/github_ci_logs.yml
still-flow: > I don't know `awk`, unfortunately.

That's what I thought, too, but it turned out much easier and elegant a solution (not that I know `awk` now, but hey):
```curl -JO $(awk -F ": " '{gsub("\r", "", $2); print $2}' response.txt)```
Also obviates your bash style nit above.

**Comment by still-flow** [comment]
Addressed all Github-related stuff. Now it remains to update Travis and write some reports to GH support (about `brew update` and log recycle period as well).

**Comment by still-flow** [comment]
_Of course_, it has only now occurred to me to [check this](https://github.community/t/adjustable-artifact-expiration/17771), but actually build artifacts also expire on a 90-day basis. Oops.

Therefore, either the new helper workflow is entirely useless and has to be nuked, or a new storage solution has to be found. Either way, this'll need some more thought.

**Comment by still-flow** [comment]
Reports to Github support both done. Will post the responses if and when they occur.

**Comment by still-flow** [comment]
Updated Travis config (loosely) in sync with the Github one. Also excised the log-saver workflow for now.

File: .travis.yml [hunk 6]
```diff
@@ -135,46 +137,51 @@ before_install:
...
 - |-
   # Install GLEW if asked for
   if [[ "$HYPERROGUE_USE_GLEW" == "1" ]]; then
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       sudo apt-get install -qq libglew-dev
-    elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      brew install glew
-    else
-      exit 'Unsupported OS'
+    elif [[ "$TRAVIS_OS_NAME" != "osx" ]]; then
```
Quuxplusone, line 152: I'd prefer to keep these if-elif ladders with all `==` conditions, so that it's easy to see which OSes are handled and which aren't. In other words, we have a case here for `linux` and a case for `osx`, and then everything else goes to the `else` case. If there's nothing to do for `osx`, then I'd just write

    elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
        # nothing to do
    else
        exit 'Unsupported OS'
    fi

(EDIT: oops, I wrote `!= "osx"` instead of `== "osx"` the first time! Evidence that keeping all the comparisons the same way around is a good idea.)

.travis/Brewfile.travis
```diff
@@ -0,0 +1,8 @@
+# using https://github.com/Homebrew/homebrew-bundle/blob/master/README.md#usage as example
+brew "sdl"
+brew "sdl_gfx"
+brew "sdl_mixer"
+brew "sdl_ttf"
+brew "glew" unless system "[[ \"$HYPERROGUE_USE_GLEW\" != \"1\" ]]"
+brew "libpng" unless system "[[ \"$HYPERROGUE_USE_PNG\" != \"1\" ]]"
+brew "automake" unless system "[[ \"$TRAVIS_USE_AUTOTOOLS\" != \"1\" ]]"
```
Quuxplusone, line 8: How is this Brewfile thing better than what came before? It seems like this is just the same list of manual `brew install` commands, except that now they're obfuscated by missing the `install` part of the `brew install`, and they're off in this new directory (which again is _not HyperRogue_ 😛). Does this Brewfile business help TravisCI cache things better or something? Do you see a speedup in build times?

.travis.yml
```diff
@@ -135,46 +137,51 @@ before_install:
...
   elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-    brew update
-    brew install sdl sdl_gfx sdl_mixer sdl_ttf
+    # SDL installation is handled by Homebrew addon above
     # work around https://stackoverflow.com/questions/51034399/ for now
     (cd /usr/local/include && ln -sf SDL/SDL.h)
   else
-    exit 'Unsupported OS'
+    echo 'Unsupported OS'
+    exit 1
```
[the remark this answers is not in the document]
still-flow: Ah, so that was intentional? I assumed it was a Travis command that got removed. I'll roll this back then.

.travis/Brewfile.travis
still-flow: Yeah, using `Brewfile` may be a bit overkill, however it's the recommended way of using Homebrew [per Travis' docs](https://docs.travis-ci.com/user/reference/osx/#homebrew). Well, the addon is, and the `bundle` deal specifically is because of the need for conditional installation.

I only really did the switch in hopes that it would work around `brew update` issues, which it didn't. So this one is liable to get rolled back as well.

still-flow:
Workflow-cancelling action doesn't seem to function properly. I guess that's another ticket for GH support right there -- asking for an official way to do so.

still-flow:
So, `podman` turned out to be slower on GH and downright not working well on Travis. Rolled that back.

In better news, forcing no `brew update` on Travis seems to have sped up OSX builds somewhat.

still-flow:
On log expiration: good news is that [Github listened](https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/). The fly in the ointment is that [for free public repos 90 days is the upper retention limit](https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy) :/. Can't blame them too much I guess: all that storage can add up quickly.

Besides possible endemic workarounds, I wanted to try out Gitlab's reportedly-celebrated CI anyways. Will report if it's any good.

zenorogue:
The CI has been broken for a long time, which was annoying. I have removed the broken stuff:
* removed TravisCI badges and YML file
* disabled testing the Windows build using gdb
* disabled testing macos-latest (there was some problem during the libraries installation due to deprecation)
* disabled Android tests (seems some problem with the build system?)
* disabled emscripten tests (actually looks like some problem with the source code, but removed anyway for now)

Also fixed some warnings which caused some Linux/Windows checks to fail.

PR #132, opened 2020-09-30
Add a CI step to compare "make clean" against "git clean"

Quuxplusone:
This supersedes #130. @still-flow, what are your thoughts now that you see it in action?

<s>@zenorogue, please don't merge this yet; at the very least I want to squash it first, and also I'd like to add the same step to the Github CI workflow as part of this PR.</s>


still-flow:
This should work well enough for PRs, where there's a natural feedback loop between committer and CI. Shame it won't help against @zenorogue (sorry, not trying to bash you here) accidentally doing the same thing, pushing the changes to Github, only to find that out later, thus necessitating a rebase -- not ideal.

(Btw, not sure if you intended this, but you can physically prevent anyone merging a PR with [this feature](https://github.blog/2019-02-14-introducing-draft-pull-requests/). It can be enabled for already created PRs [like this](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft).)

That reminds me, weren't commit hooks a thing that exists? Maybe they would be the desired "technological fix"?

Quuxplusone:
> Shame it won't help against [...] accidentally doing the same thing, pushing the changes to Github, only to find that out later, thus necessitating a rebase -- not ideal.

Not a "rebase", just a "push another commit on top that `git rm`s the accidental file." It is true that anyone who tried to base a feature branch on top of the red commit would see all their builds red until they _rebased_ their feature branch onto a green commit; but that's true of _any_ kind of red commit — bug, syntax error, whatever.

Re commit hooks, yeah, that's a technological measure that would be even further on the side of safety — "catch the issue before it's pushed, not after." I've used commit hooks at work. But client-side commit hooks have some big downsides for open-source development IMHO. They're the same as the downsides of client-side Javascript: (1) they require the client to have the right interpreter installed [bonus: we get to bikeshed over what language the commit hook should be written in], (2) they require the client to execute code controlled by the repo owner [which is a security issue], and (3) they can be bypassed by a sufficiently malicious/inept client, which means that if you want offending PRs to show up as red you're going to need a server-side check anyway. (All three of these issues are mitigated at work, where the repo owner not only trusts all the clients but also can force them to install things.) IMHO the CI step is a good idea, and IMHO it's also sufficient, for now.

still-flow:
> Not a "rebase", just a "push another commit on top that `git rm`s the accidental file."

Well, I guess that _is_ the better option in a situation like that.

> Re commit hooks

Oh, I see now. All you said does indeed sound like valid concerns. However:

1. Weren't those supposed to be just `sh` scripts? Or is choosing `sh` already a bikeshed type of thing?
2. Well, we kind of do that already with `hyperrogue` executable, although what you probably mean is that hook code is more hidden away?
3. Yeah, nature always invents a better fool :). What about doing both client-side and CI-side checks at the same time though?

Full disclosure: I've never used commit hooks before, only heard of them, so take my arguments with an appropriate amount of salt.

Quuxplusone:
> Weren't those supposed to be just `sh` scripts?

I dunno. At the place I used them, IIRC we had a `sh` script to start the process, but then it would call off into a Python script to do some of the interesting stuff. We used it to enforce `flake8`-cleanliness, "no hard tabs", alphabetization of certain arrays, etc.

> we kind of do that already with `hyperrogue` executable

And with `make` getting to execute random shell commands, too. Yes. I'll retract the security issue.

> What about doing both client-side and CI-side checks at the same time though?

That's basically my last sentence: IMHO the CI step is a good idea, [but] IMHO it's also sufficient, for now. I think it's cheap enough to test in CI that we _should_ test in CI (_Test all the things!_), but at the same time it seems rare enough not to bother growing yet another peripheral system (commit hooks) just for the purpose of catching this. I mean we've already got Makefile.simple, the autotools stuff, mymake, Travis, Github CI, and .gitignore, each of which is a thing that's _not_ HyperRogue and yet has to be maintained and bikeshedded and probably generally annoy the heck out of zenorogue. ;)

Also, I'm not even sure _how_ to write a commit hook for something that boils down to "`make clean` and then see what you deleted." We can't just run `make clean` on the client's own checkout every time they commit! The CI step, OTOH, can be really simple.

still-flow:
> each of which is a thing that's _not_ HyperRogue

Very much agreed there. Maybe I did get a bit over-enthusiastic -- I, for one, have been putting off digging into the cool math side of Hyperrogue for long enough already.

Also, sure, as a temporary solution this is much better than nothing, for all the reasons you describe. And, after all, the most permanent fixes are the temporary ones.

Quuxplusone:
Hi @zenorogue, ping! Can we commit this in order to get `hyper.html` to stop showing up as a `git diff`?

PR #134, opened 2020-10-05
Fix hyperroid builds

still-flow:
Just some simple, stupid fixes to get `hyperroid` to build again. Haven't done extensive testing, however:
- [x] game starts and starter lands are playable
- [x] upon pressing Android's "back" button and returning there's just a black (blank?) screen, bunch of `GL_INVALID_OPERATION` and `GL_INVALID_VALUE` in `adb logcat` output; force-quitting and restarting sorta solves that


still-flow:
[Here](https://github.com/still-flow/hyperrogue/actions/runs/290214559)'s a really-quickly-thrown-together test workflow run output, with APK as an added bonus (so no-one has to believe **my** word it's safe to try).

still-flow:
Re GL problem:
- apparently it exists (at least) as far back as v11.3i currently published at Play Store
- pressing "home" and returning is fine, only "back" has problems
- expanding on this, scouring the web has left me with knowing that the difference between "home" and "back" essentially boils down to the latter calling `Activity.onDestroy`, so this in conjunction with errors mentioned above makes me suspect that GL context does not get properly freed/destroyed somehow, somewhere -- especially given that force-quit fixes that

still-flow:
Side note: I've been working on adding a CI config for testing Android builds, and [here](https://github.com/still-flow/hyperrogue/actions/runs/297589836) is the result of running a more-or-less polished version of it. @zenorogue: may I interest you in adding that? And if yes, should I open a separate PR, or just expand the scope of this one?

still-flow:
@zenorogue If you've time, could you please review this? Especially the parts regarding GL bug and removed function, which you can hopefully see above (I'm not 100% certain that's how Github's review system works).

zenorogue:
Looking into Android bugs (such as the GL bug) takes some time, but I think the changes are good in general, so I am merging for now. Thanks!

The camera movement functions in rug have been changed at some point to be consistent with the implementation of HyperRogue 3D geometries; also outside of Android, it seems that the `ruggo` parameter has been commented out rather than adjusted to the new implementation. I am planning to look into this soon.

still-flow:
Aw, I sorta wanted to clean this up first -- there's even a typo in a commit message (0d174064cf4bda665e8a8670b186149ab107f209) :'(. And for some reason diff for `config.cpp` is all messed up... But I guess it's not a huge problem.

Issue #135, opened 2020-10-19
SDL_gfxPrimitives.h compilation fails as /usr/include/SDL is not in include path


akien-mga:
Compiling the latest `master` branch (as well as `11.3o` on Mageia Cauldron), I get the following error:

```
[akien@cauldron hyperrogue (master)]$ autoreconf -vif
autoreconf: Entering directory '.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force 
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: configure.ac: not using Intltool
autoreconf: configure.ac: not using Gtkdoc
autoreconf: running: /usr/bin/autoconf --force
autoreconf: configure.ac: not using Autoheader
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:7: installing './compile'
configure.ac:9: installing './config.guess'
configure.ac:9: installing './config.sub'
configure.ac:4: installing './install-sh'
configure.ac:4: installing './missing'
Makefile.am: installing './depcomp'
autoreconf: Leaving directory '.'
[akien@cauldron hyperrogue (master)]$ ./conf
config.guess  config.sub    configure     
[akien@cauldron hyperrogue (master)]$ ./configure 
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking for g++... g++
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C++... yes
checking whether g++ accepts -g... yes
checking for g++ option to enable C++11 features... none needed
checking whether make supports the include directive... yes (GNU style)
checking dependency style of g++... gcc3
checking for gcc... gcc
checking whether the compiler supports GNU C... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
checking whether gcc understands -c and -o together... yes
checking dependency style of gcc... gcc3
checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-pc-linux-gnu
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
[...]

PR #136, opened 2020-10-19
Use pkg-config in Makefile.simple.

Quuxplusone:
Let's see if TravisCI is happy with this.
Addresses, but does not fully resolve, #135.


still-flow:
Hm, it's on Ubuntu only, it seems. Additionally, plain `sdl` package seems to be discovered just fine. Could replacing `_ -> -` in other `sdl_`-packages help?

still-flow:
And sure enough, plain `libsdl1.2` provides an `sdl.pc` file, lowercase. Very consistency.

Quuxplusone:
On my OSX laptop, `pkg-config --cflags sdl SDL_mixer` works; and `pkg-config --cflags SDL sdl_mixer` also works; and `pkg-config --cflags SDL-mixer` does not work. But then, OSX's filesystem is case-insensitive by default.

Just pushed an experiment.

Quuxplusone:
There, the capitalization is sorted out now. But I'm not keen to merge this half-measure. I'd be interested to hear from @akien-mga whether this is a step in the right direction. What else (if anything) needs to be done to support `make -f Makefile.simple` on your platform?

akien-mga:
> I'd be interested to hear from @akien-mga whether this is a step in the right direction. What else (if anything) needs to be done to support `make -f Makefile.simple` on your platform?

That is a step in the right direction, but what's missing is actually setting the compile flags, and not just linker flags.
i.e. use `$(shell pkg-config --cflags sdl SDL_gfx SDL_mixer SDL_ttf)` in `CXXFLAGS_EARLY`, or a new `CXXFLAGS_SDL` you'd add to the build command.

Note that `-pthread` is a good practice to link with libpthread with GCC / Clang, but I think it should also be passed as compile flags, since it affects both the compile and linker flags: https://stackoverflow.com/questions/2127797/significance-of-pthread-flag-when-compiling

still-flow:
FWIW, I wanted to say a bit about the original issue, #135. Compare the `#include` directives in [`SDL_gfxPrimitives.h`, taken from what I believe to be upstream](https://sourceforge.net/p/sdlgfx/code/HEAD/tree/SDL_gfxPrimitives.h#l38), and in e.g. [`SDL_ttf.h`](http://hg.libsdl.org/SDL_ttf/file/703ebc7c66fd/SDL_ttf.h#l34). `<SDL.h>` in former vs. `"SDL.h"` in latter. The thing is, on all Unix distributions I'm aware of, all SDL headers lie in `SDL` subdirectory of the standard include path. So the former IIRC wouldn't work without custom `-I` switches to the compiler.

akien-mga:
@still-flow Yes, and `-I/usr/include/SDL` (or equivalent depending on where things were installed) is one of the things that the advised addition of `$(shell pkg-config --cflags sdl SDL_gfx SDL_mixer SDL_ttf)` to `CXXFLAGS` would yield.

still-flow:
Sorry, I was probably stating the obvious (:. I don't disagree, I guess I just half-heartedly tried to say that this was an upstream issue. But upstream doesn't look like it's going to be updated anytime soon, so probably this needs fixing here.

akien-mga:
It's not an upstream issue. Upstream provides you the relevant information via `pkg-config` and `sdl-config`. It's the responsibility of downstream users to set their include paths accordingly, just like you're also responsible for explicitly linking the libraries that you use.

That's why `pkg-config` exists, it provides you information *from* upstream on how to use their code downstream:
```
pkg-config --cflags --libs sdl SDL_gfx SDL_mixer SDL_ttf
-I/usr/include/SDL -D_GNU_SOURCE=1 -D_REENTRANT -lSDL_gfx -lSDL_mixer -lSDL_ttf -lSDL 
```

Quuxplusone:
My two cents: You're both (at least partly) right.

- @akien-mga is right that this PR is incomplete; it should really use `pkg-config` for the `--cflags` as well. I'm planning to update this PR, but it is low on my to-do list at the moment, and if someone wanted to suggest the exact diff, that'd be cool.

- I think @akien-mga is right that if we used `pkg-config` for both `sdl` and `SDL_gfxPrimitives`, then the upstream code in "SDL_gfxPrimitives.h" would work okay. I'm about 90% sure of this.

- @still-flow is right that "SDL_gfxPrimitives.h" currently uses the wrong kind of `#include` directive. This is a defect in the upstream code. I've known about this for years and years, having seen it in I-forget-what project. The _correct_ way for anyone to include SDL headers is `#include <SDL/whatever.h>`; that's [how everyone in the world does it](https://stackoverflow.com/questions/10488775/sdl-h-no-such-file-or-directory-found-when-compiling) _except_ for this one "SDL_gfxPrimitives.h" file. [Someone else with the same problem.](https://ubuntuforums.org/archive/index.php/t-1154208.html) I have vague recollections of emailing Andreas Schiffler about this, years ago, and getting a response _maybe_ along the lines of "these headers are no longer maintained because SDL2 makes this all obsolete." Oddly, he's in my Gmail address book, but I see no archived emails to or from; I have no evidence for my recollections.

still-flow:
Thank you both for explaining. So from all of this, it seems like the best solution would be to ditch SDL1.2 altogether, got it (;. Well, the [official migration guide](https://wiki.libsdl.org/MigrationGuide) is encouragingly not too long, so one of these days I might try going through it and seeing where it leads. (But I wouldn't suggest anyone hold their breath on that, so all of the above still applies.)

PR #138, opened 2020-11-02
Add background music volume controller to Bringris

still-flow:
Useful when you want to listen to something else, for example.

### DIFF
--- a/config.cpp
+++ b/config.cpp
@@ -1085,7 +1085,44 @@ EX void menuitem_sightrange(char c IS('c')) {
     dialog::addSelItem(XLAT("sight range"), its(sightrange_bonus), c);
   dialog::add_action(edit_sightrange);
   }
-  
+
+EX void menuitem_sfx_volume() {
+  dialog::addSelItem(XLAT("sound effects volume"), its(effvolume), 'e');
+  dialog::add_action([] {
+    dialog::editNumber(effvolume, 0, 128, 10, 60, XLAT("sound effects volume"), "");
+    dialog::numberdark = dialog::DONT_SHOW;
+    dialog::reaction = [] () {
+      #if ISANDROID
+      settingsChanged = true;
+      #endif
+      };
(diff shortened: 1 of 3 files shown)


zenorogue:
The lack of music volume setting was intentional -- the standalone Bringris includes no music, so an option to change the volume would only bring confusion.

Probably the best solution would be to check whether any music is present, and don't show the option if not (both in Bringris and in HyperRogue -- also refactor this to keep DRY.)

Also there is one legal question: since the paid versions of HyperRogue and Bringris have components which are not open source (such as the Steam interface), I would need your permission to include your changes there (otherwise it violates the GPL). What are your thoughts on this?

still-flow:
> The lack of music volume setting was intentional

Oh, I see. I really confused myself with building from main repo (:.

I agree with the second point. So, moving the relevant code into a function of its own, then, and calling it where needed?

Re legal question: I don't have any objections. Is that form of agreement sufficient for legal purposes?

still-flow:
That's the refactor, I think. I've taken the liberty to change the appearance of both volume setting dialogs to have black background in **hyperrogue** as well, because transparent BG didn't really make sense for sound settings.

zenorogue:
Thanks! I am not a lawyer, but AFAIK this is sufficient for legal purposes.

still-flow:
Oh, man, I've totally missed a regression in `emscripten` builds. There probably should be a cap guard around newly-added functions to fix that, if having cap guards in two different places is OK. Maybe moving these into function bodies would be better?

zenorogue:
I have added cap guards inside the functions, so the function can be safely called if no audio is available, they will simply not display anything. 

Also disabled the music volume item not on `!audio`, but on the newly added flag `!music_available` (Bringris includes an empty `hyperrogue-music.txt` file which sets audio to true so that sound effects are played, but it will not trigger `music_available` unless some music is actually defined).

PR #142, opened 2021-03-11
Eliminate autotools/automake/configure.

Quuxplusone:
IMO, the autotools stuff has outlived its usefulness. Its "version number" is never up-to-date, and it currently gives a bunch of warnings on Linux because its warning options are also out-of-date. Plus, it's just more stuff to test in CI; and it makes the `.gitignore` full of random files; and it's 534 lines of stuff-that's-not-really-HyperRogue. I say delete it.


zenorogue:
I agree, but let's wait a bit to see if anybody has any objections.

zenorogue:
Should we also rename `Makefile.simple` to just `Makefile` while we are at it? This will also help to fix the issue of being unable to compile mymake with `make mymake`.

Quuxplusone:
> Should we also rename `Makefile.simple` to just `Makefile` while we are at it?

Sure, I'll update the PR (and rebase to poke the CI, too).

Quuxplusone:
Updated; and I've dumped 1761b02 in here as well, to try to fix https://travis-ci.org/github/zenorogue/hyperrogue/jobs/763799466 .

Quuxplusone:
(I wonder whether removing the gdb stuff from `.github/workflows/test_simple.sh` would fix https://github.com/zenorogue/hyperrogue/pull/142/checks?check_run_id=2161586673 , but I'll leave that experiment out of this PR.)

Quuxplusone:
:tada:

PR #153, opened 2021-06-09
Fix typos in Russian translation.

blargdag:
Fixed a few minor typos.


blargdag:
@zenorogue  First time contributor here. How do I get approved for the workflows?

zenorogue:
Thanks!

This "approved for the workflows" seems to be about Github CI which is still broken, so it is irrelevant anyway.

PR #163, opened 2021-06-27
The Orb of Beauty now lifts rose-scent restrictions on targeting of ranged orbs

jruderman:


zenorogue:
Thanks for your pull requests!

One legal question: since the paid versions of HyperRogue have components which are not open source (such as the Steam interface), I need your permission to include your changes there (otherwise it violates the GPL). What are your thoughts on this?

jruderman:
Yes, that's fine. You have my permission to include my changes to HyperRogue in versions that include non-open-source components.

PR #170, opened 2021-06-29
Remove extra semis in struct defs

jruderman:
These were found using clang's -Wextra-semi﻿


Quuxplusone:
Could you also add `-Wextra-semi` to the Makefile, in the obvious place for Clang?
As long as the TravisCI build is still happy, that is.
(The GithubCI ubuntu build appears to be broken right now; I don't know what the deal is with that. Also, the TravisCI emscripten build [is broken](https://travis-ci.org/github/zenorogue/hyperrogue/jobs/774012122) with a very weird symptom; anyone know what's up?)

PR #171, opened 2021-06-29
Remove unused catch binding

jruderman:
If I enable -Wunused-exception-parameter, this instance is the only warning I get when compiling HyperRogue.﻿


Quuxplusone:
Could you also add `-Wunused-exception-parameter` to the Makefile, in the obvious place for Clang?

PR #173, opened 2021-07-01
Orb of Safety excludes more monsters

jruderman:
Based on a report from 'Shield' on Discord, I investigated which monsters can appear in the initial field when using an Orb of Safety.

Changed to exclude on safety:
* Frogs (in Frog Park)
* Ivies, Giant Apes, Eagles (in Lost Mountain, frequently visible from Jungle)
* Red Raiders (in Ruined City)

Unchanged and can still appear on safety:
* Parrots in Caribbean (they can't chase the player into the sea)
* Vine Spirits in Vineyard (they help prevent stalemate)
* Volcanic Wasteland salamanders and Irradiated Field monsters (these lands don't have orbs of safety, but can appear occasionally at the periphery when using safety in another land)
* Reptile floors can still be about to awaken

### DIFF
--- a/landgen.cpp
+++ b/landgen.cpp
@@ -2192,11 +2192,11 @@ EX void giantLandSwitch(cell *c, int d, cell *from) {
     
     case laMountain:
       if(d == 7) {
-        if(hrand_monster(50000) < 100)
+        if(hrand_monster(50000) < 100 && !safety)
           buildIvy(c, 0, 3);
-        else if(hrand_monster(125000) < 100 - celldistAlt(c))
+        else if(hrand_monster(125000) < 100 - celldistAlt(c) && !safety)
           c->monst = moMonkey;
-        else if(hrand_monster(200000) < min(100, -10*celldistAlt(c)) - celldistAlt(c))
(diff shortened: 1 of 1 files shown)


Quuxplusone:
Scope-creep idea: I haven't studied this code, but the commit summary makes it sound as if it's a big pain to decide for each monster whether the Orb should affect it because you're disallowing a monster _only_ if it appears in the same or adjacent lands as an Orb. If there's any change to the set of lands-in-which-Orb-of-Safety-appears, or any change to the possible-adjacencies-between-lands, or any change to the set of monsters-that-can-appear-in-those-lands, then your hard work is invalidated.

What if there were just a master list of "dangerous monsters," and the Orb would prevent dangerous monsters from being placed, period?

To mark a monster as "dangerous," we could use a flag on the monster record itself, or just have a `bool isDangerous(eMonster)` function with a big `switch` in it.
To prevent placing monsters, we could just replace e.g.
```
c->monst = moMonkey;
```
with
```
c->spawnMonster(moMonkey);

void cell::spawnMonster(eMonster m) {
  if (isDangerous(m) && safety) return;
  c->monst = m;
  }
```

PR #176, opened 2021-07-03
Orb of the Dragon can no longer target underwater items

jruderman:
This fixes a gameplay exploit where shift+hovering with Orb of the Dragon revealed which sea cells contain hidden items.﻿


zenorogue:
I think that makes it impossible to burn items on unburnable terrain.
The logical thing IMO is to make water prevent items for burning -- so done it that way.

PR #190, opened 2021-07-08
Fill in some missing Russian translations in menus

blargdag:
Also some minor translation fixes.


zenorogue:
Thanks! It is good if the translation files are in the same order, so if a translator wants to update the translation to the current version, they can simply see where the Russian translation ends, find the same place e.g. in the Polish translation, and proceed from there. Adding "restructured menus" is helpful, but it breaks this and could be confusing, so I would add a line of comment about this. (e.g. "//note: the translation should be complete until the marked line" and "// this is the marked line")

blargdag:
Hmm OK, I'll see if I can rearrange the stuff.  Problem is, I'm not really translating everything; how to indicate which strings have been skipped over? Copy from e.g., the Polish translation and comment them out?

blargdag:
Also, just realized some of the translated items are not yet present in any other translation files. How to indicate this? Append to the end of the file?

zenorogue:
I think just adding the comments like I have written above would be clear and fine. Not sure what you are referring to in your last comment.

blargdag:
I'm still confused by what you said about keeping the same order in the translation files. What I have currently are translations of isolated strings (e.g., in menus where untranslated strings are prominent), but they're not necessarily in the order they would appear in, e.g., the Polish translation file.  So what's the right way to deal with this?  Just append them to the end of the file? I'm not sure where your suggested comments should be put?

Or copy the Polish translations up to the line corresponding with my last translation, but comment out the lines I don't have translations for, to indicate they are not yet translated? E.g.

````
S(..., ...) // last translated string
//S(..., ...) // copied from Polish translation, commented out
//...
S(..., ...) // available translation
//S(...,...) // more from Polish translation, commented out
````

blargdag:
My second comment was referring to strings I translated that haven't been translated into, e.g., Polish yet.  Where should these strings be put, keeping in mind that we want to keep the same order between translation files?  I guess at the end of the file, but there remains a large amount of strings that do have Polish translations, but don't have Russian translations yet.  How to reconcile this?

blargdag:
P.S. Split off the non-order-dependent fixes into a separate PR while we hash out this one.

zenorogue:
I meant as follows:

Polish:
```A
B
C
D
E
F```

Russian:
```A
B
C
// this is the marked line
E
//note: the translation should be complete until the marked line```

blargdag:
Ah, gotcha. Will do that.

blargdag:
Done, also added a bunch of (partial) translations for various lands.

PR #193, opened 2021-07-10
Orb of shield/shell no longer bypasses 'kraken will destroy boat' warning

jruderman:


jruderman:
Oops, I need to resubmit this without it being from 'jruderman/master'

PR #212, opened 2021-07-10
Prevent land adjacencies with fire danger

jruderman:
Some of these are extremely dangerous (r'lyeh + rose). Most of them are at least somewhat dangerous and weird thematically.

### DIFF
--- a/landlock.cpp
+++ b/landlock.cpp
@@ -248,17 +248,19 @@ EX int isRandland(eLand l) {
   return 0;
   }
 
+EX bool landStartsFires(eLand l) {
+  return (l == laPower || l == laVolcano || l == laDragon || l == laEFire || l == laRlyeh);
+  }
+
+EX bool landCatchesFire(eLand l) {
+  return (l == laWineyard || l == laDryForest || l == laEndorian || l == laRose);
+  }
+
 EX bool incompatible1(eLand l1, eLand l2) {
   if(isCrossroads(l1) && isCrossroads(l2)) return true;
+  if(landStartsFires(l1) && landCatchesFire(l2)) return true;
   if(l1 == laJungle && l2 == laMotion) return true;
   if(l1 == laMirrorOld && l2 == laMotion) return true;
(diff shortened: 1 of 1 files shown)


Quuxplusone:
I (think I) see what you mean about how if the Rose Garden catches fire, you probably just die. And I definitely think your patch cleans up the intent of the code very well. But... maybe the Dry Forest should be moved from the "catches fire" list to the "starts fires" list? Because:

- Dry Forest *does* catch fire, but it's not catastrophic when it does, right? It's practically expected that it will.
- I think it'd be pretty cool if you managed to start a fire in the Dry Forest by chaining from the Land of Power, so let's keep that coolness possible.
- But it should be added to the other list, because a fire in the Rose Garden is equally dangerous no matter whether it's started by a Fire Cultist (R'lyeh) or a Fire Fairy (Dry Forest).

In other words, I think the lists should be "has any way to start fires" and "is _especially unbalanced_ if it ever catches fire." The latter list wouldn't contain Dry Forest, and might not contain some other lands too, I dunno.

jruderman:
Maybe the Dry Forest should be on both lists? Getting Dragon'd in the Dry Forest is a dangerous surprise, but having a Fire Fairy follow you into Yendorian Forest is also dangerous.

ammongit:
I think it should be on both lists, it's both a source of fires and pretty radically changes when it catches fire (even if it's expected that it will).

landlock.cpp
```diff
@@ -248,17 +248,19 @@ EX int isRandland(eLand l) {
   return 0;
   }
 
+EX bool landStartsFires(eLand l) {
+  return (l == laPower || l == laVolcano || l == laDragon || l == laEFire || l == laRlyeh);
```
ammongit, line 252: As discussed in comments:
```suggestion
  return (l == laPower || l == laVolcano || l == laDryForest || l == laDragon || l == laEFire || l == laRlyeh);

```

PR #213, opened 2021-07-10
Move land check functions to the top

ammongit:
Minor PR that is a complement to https://github.com/zenorogue/hyperrogue/pull/212.

It standardizes the function such that land check functions appear at the top rather than scattered throughout.


jruderman:
Could these symmetric checks (crossroads/elemental) be moved from incompatible1() to incompatible()?

ammongit:
I see, `incompatible()` is the wrapper. You're suggesting moving it because the symmetry means additional checks are being performed that are not necessary.

PR #214, opened 2021-07-10
Strip trailing whitespace

emmiegit:
While editing the files, I noticed a lot of them had trailing whitespace without any particular reason for being there. This PR removes them.

You can verify this only affects whitespace using the following setting in the "files" tab:
image


zenorogue:
Thanks for this PR! I have not accepted it because IMO the disadvantages seem to be greater (the possibility of breaking patches because the patcher is not sure to do because whitespace does not match). Similar with another pull request (#283). However, I usually try to avoid trailing whitespaces in new commits.

PR #220, opened 2021-07-11
Use offsetof to reduce UB

jruderman:
During startup, tailored_alloc() trips Clang's UBSan:

```
locations.cpp:184:28: runtime error: member access within null pointer of type 'const hr::heptagon'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior locations.cpp:184:28 in
```

and so does connection_table's spintable():

```
locations.cpp:137:73: runtime error: member access within address 0x60c000064278 which does not point to an object of type 'hr::heptagon'
0x60c000064278: note: object has invalid vptr
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior locations.cpp:137:73 in
```

Both of these can be solved by switching from fake-pointer-ops to offsetof, which is intended for this purpose. Unfortunately, using offsetof with these types triggers a warning at compile time:

```
locations.cpp:183:11: error: offset of on non-standard-layout type 'hr::cell' [-Werror,-Winvalid-offsetof]
```

The warning can be suppressed for Clang using -Wno-invalid-offsetof. I think this is legit: [offsetof is widely supported beyond 'standard-layout classes', and future C++ standards may remove the warning in some or all cases](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0897r0.html).

I have only tested this with Clang on macOS.


locations.cpp
```diff
@@ -178,10 +178,9 @@ template<class T> struct connection_table {
...
 
 template<class T> T* tailored_alloc(int degree) {
-  const T* sample = nullptr;
   T* result;
 #ifndef NO_TAILORED_ALLOC
-  int b = (char*)&sample->c.move_table[degree] + degree - (char*) sample;
+  int b = offsetof(T, c) + offsetof(connection_table<T>, move_table) + sizeof(T*) * degree + degree;
   result = (T*) new char[b];
   new (result) T();
 #else
```
Quuxplusone, line 186: I'd rather see
```
#ifdef NO_TAILORED_ALLOC
  result = new T;
#else
  int b = offsetof(T, c.move_table[degree]) + degree;
  result = ::new (::operator new(b)) T;
#endif
```
and in `tailored_delete`,
```
#ifdef NO_TAILORED_ALLOC
  delete x;
#else
  x->~T();
  ::operator delete((void*)x);
#endif
```

mymake.cpp
```diff
@@ -75,7 +75,7 @@ void set_linux() {
 
 void set_mac() {
   preprocessor = "g++ -E";
-  compiler = "g++ -march=native -W -Wall -Wextra -pedantic -Wno-unused-parameter -Wno-implicit-fallthrough -c";
+  compiler = "g++ -march=native -W -Wall -Wextra -pedantic -Wno-unused-parameter -Wno-implicit-fallthrough -Wno-invalid-offsetof -c";
```
Quuxplusone, line 78: Off-topic: I notice a bunch of new `-Winconsistent-missing-override`, `-Woverloaded-virtual`, and `-Wextra-semi` warnings locally on OSX. We should probably fix those somewhere. :)

PR #239, opened 2021-07-15
Peace mode: don't switch places with dice

jruderman:
This prevents a crash due to the dice map not having an entry for the Animated Die's new location.

However, it allows the player to occupy the same cell as an Animated Die, which is weird, so I'm not sure this is the best solution.


ammongit:
Is it possible to turn dice enemies into movement restrictions on peaceful? "You cannot move through the Angry Die!" type thing?

zenorogue:
Trying to move into a die, multi-tile monster, or a saved tortoise now generates a message saying this is impossible. (Also fixed the lack of message when no place to switch with a Golem.)

PR #246, opened 2021-07-18
Adc

### LINKED ISSUE #244: ADC implicitly adds empty entries to all_drawn_copies
Autoplay in Dry Forest spends a lot of time in drawFlashes, and part of the reason is the way `ADC` accesses `all_drawn_copies`.

**Current code**

```
#define ADC(V,c) for(const shiftmatrix& V: current_display->all_drawn_copies[c])
```

When `all_drawn_copies` lacks an entry for `c`, this _implicitly creates an entry_ with an empty vector and inserts it into the map.

**Possible solution 1: check map::count first**

```
#define ADC(V,c) if (current_display->all_drawn_copies.count(c)) for(const shiftmatrix& V: current_display->all_drawn_copies[c])
```

This has the disadvantage of searching the map twice: once for `count` (or `contains`), once for `[]` (or `at`). Still, it seems faster overall.

**Possible solution 2: create an outer loop using map::find**

```
#define ADC_helper(V,c,adc) for(auto it = adc.find(c); it != adc.end(); it = adc.end()) for(const shiftmatrix& V: it->second)
#define ADC(V,c) ADC_helper(V, c, current_display->all_drawn_copies)
```

This only searches the map once, and avoids modifying the map, but there's a lot about map iterators that I don't understand.

* Does this make `it` contain its own copy of the `vector<shiftmatrix>`? If so, maybe `const auto &it` could avoid copying, but that made Clang balk at `it = adc.end()`.
* Does `adc.end()` contain or point to a fake element with an empty vector? If so, where do these come from?

Quuxplusone:
Fixes #244.

Right now the replacement code is of the form
```
for (const shiftmatrix& V : hr::span_at(current_display->all_drawn_copies, dragon::target)) {
```
but maybe it would be nicer to make it a member function of `current_display`; what do you think?
```
for (const shiftmatrix& V : current_display->all_drawn_copies_at(dragon::target)) {
```

I didn't do that mainly because I don't know whether the preposition should be `_at` or something else. I just refactored the C++, didn't really try to understand what it _means_. ;)

### DIFF
--- a/graph.cpp
+++ b/graph.cpp
@@ -45,7 +45,24 @@ EX bool hide_player() {
      ;
   }
 
-#define ADC(V,c) if (auto *it = hr::find_or_null(current_display->all_drawn_copies, c)) for(const shiftmatrix& V: it->second)
+template<class T>
+class span {
+  T *begin_ = nullptr;
+  T *end_ = nullptr;
+
+  public:
+  explicit span() = default;
+  explicit span(T *p, int n) : begin_(p), end_(p + n) {}
+  T *begin() const { return begin_; }
+  T *end() const { return end_; }
+  };
+
+template<class Map, class Key>
+hr::span<const shiftmatrix> span_at(const Map& map, const Key& key) {
+  auto it = map.find(key);
+  return (it == map.end()) ? hr::span<const shiftmatrix>() : hr::span<const shiftmatrix>(it->second.data(), it->second.size());
+  }
+#define ADC(V,c) for (const shiftmatrix& V : hr::span_at(current_display->all_drawn_copies, c))
 
 EX hookset<bool(int sym, int uni)> hooks_handleKey;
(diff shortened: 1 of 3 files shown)


zenorogue:
74052ef is C++17, isn't it?  (The Steam compiler is still stuck at C++11)

Quuxplusone:
@zenorogue wrote:
> 74052ef is C++17, isn't it? (The Steam compiler is still stuck at C++11)

Nope! Actually, my original code used some `auto` return types (C++14), and the Makefile forcibly reminded me to stick to `-std=c++11`. :)  You're thinking of C++17 `if (auto x = y; z)`, but simple `if (auto x = y)` has been legal since C++11 (and `if (T x = y)` since C++98). Relevant blog post: https://quuxplusone.github.io/blog/2020/10/28/while-with-initializer/

PR #261, opened 2021-07-21
Change goblin text to account for its presence in other lands

ammongit:
The message for the Goblin enemy currently appears as:
screenshot of goblin's flavor text

However Goblins also appear in the Dead Cave (as shown in the screenshot). This PR implements my suggestion from the HyperRogue discord to replace "Living Cave" with "caves" more generally.

### DIFF
--- a/language-de.cpp
+++ b/language-de.cpp
@@ -641,8 +641,8 @@ S("Rangers take care of the magic mirrors in the Land of Mirrors. "
   "Ranger beschützen die Zauberspiegel im Spiegelland. Sie wissen, dass Schurken wie du "
   "die Spiegel gerne zerbrechen - also greifen sie dich an!")
 
-S("A nasty creature that lives in caves. They don't like you "
-  "for some reason.",
+// TODO update translation
+S("A nasty creature that lives in caves. They don't like you for some reason.",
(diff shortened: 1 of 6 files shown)


language-de.cpp
```diff
@@ -641,7 +641,7 @@ S("Rangers take care of the magic mirrors in the Land of Mirrors. "
   "Ranger beschützen die Zauberspiegel im Spiegelland. Sie wissen, dass Schurken wie du "
   "die Spiegel gerne zerbrechen - also greifen sie dich an!")
 
-S("A nasty creature native to the Living Caves. They don't like you "
+S("A nasty creature that lives in caves. They don't like you "
   "for some reason.",
   "Ein grässliches Geschöpf, das in den Lebenden Höhlen haust. Irgendwie können diese Kreaturen dich nicht leiden.")
```
Quuxplusone, line 646: Nit 1: You could join the string literals on lines 644 and 645 together.
Nit 2: In #111, I copied from #154 the convention of putting `//TODO FIXME` at the end of every line where the English has changed but the other language hasn't been updated yet. I think you should do so in this PR, as well.

ammongit: Added both of these.

PR #271, opened 2021-08-02
Adjust Snake Nest colors so drawMonster can add them without overflowing to black

jruderman:
drawMonster has:

```
      int c1 = nestcolors[pattern_threecolor(c)];
      int c2 = nestcolors[pattern_threecolor(c->move(d))];
      col = (c1 + c2); // sum works because they are dark and should be brightened
```

So, the colors in nestcolors need to have each color component below 0x80 in order for the snakes not appear black.

### DIFF
--- a/graph.cpp
+++ b/graph.cpp
@@ -3362,7 +3362,7 @@ ld wavefun(ld x) {
   else return 0; */
   }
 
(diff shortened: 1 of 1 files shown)


jruderman:
Or maybe it's supposed to be impossible for Snake Nest snakes to be entirely on a single color? Currently they can spawn as red-green alternating but switch to green-green as they pursue the player.

PR #284, opened 2021-08-04
Stop setting landparam in Hunting Ground generation

jruderman:
* Seems to be unused: I didn't see any code related to Hunting Ground or dogs that checks landparam.
* Could leak into other lands: for example, with `./hyper -C -W Hunting -W2 Minefield -I Turquoise 1`, some dogs that generate on Minefield cells would also incorrectly mark their Minefield cell as "known mine".


zenorogue:
Seems unused indeed (maybe some old debug?) but I added checks `if(dog2->land == laHunting)` for safety.

PR #292, opened 2021-08-12
More Russian translations

blargdag:
Translate descriptions of Brown Islands, Free Fall.
Advance marked line. :-)


zenorogue:
The Polish translation file https://github.com/zenorogue/hyperrogue/blob/master/language-pl.cpp is updated with some new texts, so it would be good to update the French translation too.

PR #299, opened 2021-11-17
Add HYPERFONTPATH to make it easier to package on Fedora.

dulsi:
Fedora prefers not to have multiple copies of fonts. Since the font is already installed, I need to specify a different location than the other game files. HYPERFONTPATH defaults to the same as HYPERPATH so no one needs to use it unless they are in a similar situation.

Linux has a fontconfig library that can be used to lookup a font instead of hardcoding the path. Would you be interested in a patch to add that capability to the Linux version. I used it in ostrichriders:

https://github.com/dulsi/ostrichriders/commit/0862a5c8b7f74c12c5ff931511d53cfc43ca4f66

### DIFF
--- a/basegraph.cpp
+++ b/basegraph.cpp
@@ -197,7 +197,7 @@ EX void present_screen() {
 
 #if CAP_SDLTTF
 
-EX string fontpath = ISWEB ? "sans-serif" : HYPERPATH "DejaVuSans-Bold.ttf";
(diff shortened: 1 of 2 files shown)


zenorogue:
Thanks!

A capability to lookup a font using fontconfig sounds useful, although I would prefer this to be an optional dependency.

PR #303, opened 2022-03-08
Extra canvas-specific options + extra animation options

lottieratworld:
### Extra canvas options
Added new options in the map editor to change the floor shape and the darkening of the canvas land

### Extra animation options
New "idle animation" option (wasn't sure of the best name to give it) that changes the speed of animations such as item rotation, orb effects, color shifting, etc.
New "flashing effects" toggle that, when toggled off, replaces effects such as lightning and the orb effect animations of the Orb of Storms, Curse of Repulsion, and Curse of Gluttony with slower smooth animations that can be adjusted by the above "idle animation" option


zenorogue:
Thanks! One legal question: since the paid versions of HyperRogue (and potential future games using this engine, such as Bringris) have components which are not open source (such as the Steam interface), I would need your permission to include your changes there (otherwise it violates the GPL). What are your thoughts on this?

lottieratworld:
of course go ahead. perfectly fine by me :)

PR #319, opened 2022-09-17
Provide orbs with shapes that hint at their function

jlmjlm:
I've found it impossible to keep track of which orbs are which due to them all having the same disk shape that's differentiated only by color and their rings only distinguishing major orb classes.  This PR adds icon shapes to the bulk of orbs to help identify them visually.  Often this is clear (eg, a golem in the Orb of Life and a lightning bolt in the Orb of Storms), but often it's more abstract and hence dubious but still worth a try.  I've attempted to mostly preserve coloration to not be disruptive to those who have achieved automatic color-based orb identification.

This is my first dive into hyperrogue's source code.  I've attempted to honor the existing code style, but I've probably overlooked some aspects.  I've also probably overlooked some existing utility routines and organization conventions.  Please point such oversights out!

Attached are old & new screenshots of an "all orbs" level to show the visual result of the changes.
Old: orbs-disks
New: 
orbs-shapes


zenorogue:
Looks awesome! People like this very much in general, although some find some of them questionable (see Discord). So the plan is to merge this pull request, make this an option (turned on by default), and modify the questionable ones a bit. But first one legal question: since the paid versions of HyperRogue (and potentially other RogueViz games) have components which are not open source (such as the Steam interface), I would need your permission to include your changes there (otherwise it violates the GPL). What are your thoughts on this?

jlmjlm:
Many of my shape choices can be reasonably questioned, for sure, and I'm open to suggestions!  

And I'm fine with this being used in the non-GPL versions.  Do you have a form for contributors to waive copyright, or does this statement suffice?

zenorogue:
This statement should suffice, thanks!

PR #320, opened 2022-10-05
Flip over worm-head shape used in Orb of Domination.

jlmjlm:
Worms get drawn from tail to head, so the shape used for the head is upside-down.  Hence it should be flipped over when used as the icon for the Orb of Domination.  (Close and ignore if having the icon upside-down was a deliberate aesthetic choice.)


zenorogue:
I imagined the worm rising its head from the sand and thus heaving its teeth on top. But I have no strong opinion of this, so pulled anyway.

PR #324, opened 2022-11-04
Correct URL in help text.

jlmjlm:


help.cpp
```diff
@@ -185,7 +185,7 @@ EX void buildHelpText() {
 #endif
 #endif
   help += XLAT("See more on the website: ") 
-    + "http//roguetemple.com/z/hyper/\n\n";
+    + "https://roguetemple.com/z/hyper/\n\n";
```
Quuxplusone, line 188: Adding the missing `:` seems good, but presumably if `http` is going to become `https` here then it should become `https` throughout the codebase. `git grep -o http.*roguetemple` shows no `https`es at the moment.

PR #327, opened 2022-11-11
OSM: Make the orb quantity in the inventory screen black.

jlmjlm:
An unintended side effect of the orb icons I created was to reduce the legibility of the orb quantities in OSM, because they also operate by darkening a shape (in this case, a numeral instead of an icon) on the orb's disk.

This aims to fix that problem I introduced by making the number overlay black, which should be a good contrast with any orb.

Before:
osm_before

After:
osm_after

### DIFF
--- a/inventory.cpp
+++ b/inventory.cpp
@@ -547,10 +547,8 @@ EX namespace inv {
                 quickqueue();
                 }
               
-              int tcol = remaining[i] ? darkenedby(icol, 1) : 0;
-  
               if(remaining[i] != 1 || !gg)
-                displaystr(px, py, 2, gg?rad:rad*3/2, remaining[i] <= 0 ? "X" : remaining[i] == 1 ? "o" : its(remaining[i]), tcol, 8);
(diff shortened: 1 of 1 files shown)


Quuxplusone:
I wonder: I've never looked at the graphics-display code, but is it possible to do "black text with a white outline" (or vice versa), and if so, maybe do that and even move the number to the lower right corner of the orb so it doesn't overlay the identifying icon in the middle of the orb?  Or is that basically asking for the impossible? :)

jlmjlm:
Please feel free to make alternative or conflicting PR's, or use this one as a base to modify, if you think yours will look better.  It's entirely possible I'll agree!

zenorogue:
It is of course possible to make black text with a white outline, but HyperRogue has "default dialog text color" and "default boundary color" settings, and there is no good reason to ignore these settings in this context.

The idea to move the number to the lower right corner is great though.

So now it uses the old style if the orb mode is not "icons", and a default-dialog-colored number in bottom left if the orb mode is "icons".

The changes are already pushed. Thanks for the ideas!

jlmjlm:
I just did an OSM run, and I like the look of what you did instead, Zeno.  Thanks!

PR #330, opened 2022-12-28
Reset hr::instat in drawStats() if cornermode is enabled.

jlmjlm:
This appears to fix a bug where the HUD's north-of-map text gets stuck in "Press F1 or right click for help".

To reproduce:
* Narrow the window so that the treasures/monsters/orbs HUD indications switch from two-sides to four-corners.
* Mouse over a monster HUD indicator in the northeast area.
* Mouse over the map.

Expected behavior is that the description text on the HUD's north edge gives descriptions of the land.

The behavior appears to be correct after this is applied, but this is the first time I've investigated the logic setting and resetting `hr::instat`, so I'm not positive that it should always be reset here.

### DIFF
--- a/hud.cpp
+++ b/hud.cpp
@@ -476,13 +476,13 @@ EX void drawStats() {
     quickqueue();
     }
   
-  if(racing::on) 
+  if(racing::on) {
 #if CAP_RACING
     racing::drawStats();
-#else
-    {}
 #endif
+    }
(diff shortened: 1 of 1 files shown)


zenorogue:
Thanks for the diagnosis! I believe I have used it to write an independent fix (not noticing the PR), so this should be closed.

PR #333, opened 2023-02-01
Make it build on linux + various fixes

iacore:
The `EX` macro is messing with the code, so I deleted one.

Everything else should be self-explanatory.


hyperpoint.cpp
```diff
@@ -21,8 +21,8 @@ namespace hr {
 static constexpr ld A_PI = M_PI;
 static constexpr ld TAU = 2 * A_PI;
 static constexpr ld degree = A_PI / 180;
-static constexpr ld golden_phi = (sqrt(5)+1)/2;
-static constexpr ld log_golden_phi = log(golden_phi);
+static constexpr ld golden_phi = noexcept((sqrt(5)+1)/2);
+static constexpr ld log_golden_phi = noexcept(log(golden_phi));
```
Quuxplusone, line 25: Probably obvious to @zenorogue, but just in case: obviously don't merge this change, unless you're OK with `golden_phi` becoming `0.0`. 😛

Quuxplusone (changes requested):

hyperpoint.cpp
iacore: oh wait

iacore:
@zenorogue can you take a look at this?

zenorogue:
Everything looks good. Merged, thanks!

PR #342, opened 2023-08-23
Fix -Wunused-but-set-variable warnings from Clang

Quuxplusone:
You should take a close look at each of these removals, because they _might_ indicate a typo further down — maybe the variable was intended to be used, but accidentally wasn't. For example, `return a < 6;` is very suspicious: should it have been `return a0 < 6;` ?


barriers.cpp
```diff
@@ -1039,8 +1039,6 @@ EX bool valid_dir(const vector<char>& ad, int j, cell *c) {
   int a = 0;
   for(auto& dd: ad1) if(dd == 1) a++;
 
-  int a0 = 0;
-  for(auto& dd: ad) if(dd == 1) a0++;
   return a < 6;
```
Quuxplusone, line 1042: This one still seems suspect to me.

PR #361, opened 2024-02-21
Record World Turtle kills.

jlmjlm:
Clean up "fake" monsters:
* moNone is used for melted walls.
* moIvyNext is actually tracked.
* moWolfMoved is newly used for World Turtles.
* moREMOVED isn't used.


zenorogue:
Thanks! Merged. "Fake" means here that the value should not be displayed in the local score list; so disabling "fake" for moNone caused the score list to have a confusing column called "no monster". I have fixed this so that this column is accurately named "icewalls melted".

PR #372, opened 2024-04-22
Check tide before drawing cells.

jlmjlm:
Fixes bug where on the first turn in a plane, the ocean cells are all displayed as water at high tide and volcano cells are all displayed as lava at low tide.

### DIFF
--- a/celldrawer.cpp
+++ b/celldrawer.cpp
@@ -2830,6 +2830,8 @@ void celldrawer::draw() {
 
   cells_drawn++;
 
+  checkTide(c);
+
 #if CAP_TEXTURE
   if(texture::saving) {
     texture::config.apply(c, V, 0xFFFFFFFF);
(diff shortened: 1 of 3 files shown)


jlmjlm:
The bug this fixes is still present in 13.0h.  What's your plan for handling it?

https://github.com/zenorogue/hyperrogue/assets/501357/9d122f0b-b564-4567-b863-b8053ac819dd

zenorogue:
Well, `checkTide` was called in `celldrawer::draw` later, but only under some conditions. Calling `checkTide` might not be recommended -- in the chaos mode, tides use a special implementation, which might need several iterations to converge, and also potentially be quite slow. But yeah, probably we should ignore that and always simply call `checkTide`. Thanks!

jlmjlm:
If redundant calls to `checkTide` is a problem, I'd hope that some form of memoization could address that and allow the expensive computation to happen once per move.

PR #375, opened 2024-05-16
Make option to show turns on the main screen.

jlmjlm:
"Liberal" version that shows the turncount unguarded by nomenukey and ISMOBILE.  Hopefully this means that it'll display properly on mobile, outside the map region's "DRAG" area. But this display mode is only used by the mobile builds which I don't have toolchains for, so can't test it.  :(

### DIFF
--- a/config.cpp
+++ b/config.cpp
@@ -1173,6 +1173,8 @@ EX void initConfig() {
 
   param_i(min_cells_drawn, "min_cells_drawn");
 
+  param_b(show_turns, "show_turns", false)
+  -> editable("show turn count", 'T');
   param_i(menu_darkening, "menu_darkening", 2)
   -> editable(0, 8, 1, "menu map darkening", "A larger number means darker game map in the background. Set to 8 to disable the background.", 'd')
   -> set_sets([] { dialog::bound_low(0); dialog::bound_up(8); dialog::get_di().dialogflags |= sm::DARKEN; });
@@ -2443,6 +2445,7 @@ EX void configureInterface() {
       };
     });
 
+  add_edit(show_turns);
   add_edit(menu_darkening);
   add_edit(centered_menus);
   add_edit(startanims::enabled);
--- a/graph.cpp
+++ b/graph.cpp
@@ -5672,6 +5672,7 @@ EX bool just_refreshing;
 
 EX int menu_darkening = 2;
 EX bool centered_menus = false;
+EX bool show_turns = false;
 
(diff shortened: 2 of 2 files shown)


zenorogue:
Thanks!

For testing the mobile-style interface, there is fake-mobile, compiled with e.g.
`g++ fake-mobile.cpp -o fake-mobile -I/usr/include/SDL -lSDL -lSDL_gfx -lGL -lSDL_ttf -lz`

jlmjlm:
Neat!  Unfortunately, I persistently get an immediate segfault when I try to run `fake-mobile`:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7e58740 in TTF_SizeUNICODE () from /lib/x86_64-linux-gnu/libSDL_ttf-2.0.so.0
(gdb) bt
#0  0x00007ffff7e58740 in TTF_SizeUNICODE ()
   from /lib/x86_64-linux-gnu/libSDL_ttf-2.0.so.0
#1  0x00007ffff7e58af5 in TTF_SizeUTF8 ()
   from /lib/x86_64-linux-gnu/libSDL_ttf-2.0.so.0
#2  0x0000555555a9b982 in hr::textwidth(int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
#3  0x0000555555901041 in hr::displayfr(int, int, int, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int, int) ()
#4  0x00005555559d5216 in hr::displayabutton(int, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int) ()
#5  0x0000555555a40f3f in hr::gamescreen() ()
#6  0x0000555555a41897 in hr::normalscreen() ()
#7  0x0000555555ec19dd in hr::function_state<void (*)(), void>::call() const ()
#8  0x0000555555be36cb in hr::function<void ()>::operator()() const ()
#9  0x0000555555a41e71 in hr::drawscreen() ()
#10 0x0000555555a9a5e5 in hr::mobile_draw(int) ()
#11 0x0000555555a9bbc9 in main ()
```

This persists even if I change the guard on the turncount condition to `show_turns && false`, which makes sense because `gamescreen()` is called shortly before the turncount would be displayed.

zenorogue:
(For the record: this was caused by incorrect font file.)

I am going to replace this with a programmable display, so the player can choose the parameter to be displayed there.

PR #376, opened 2024-05-16
Remove trailing whitespace from system.cpp

jlmjlm:


jlmjlm:
This is a _whitespace only_ change.
That CI fails this PR means that CI is broken.

zenorogue:
Thanks! Fixed these warnings.

PR #378, opened 2024-05-17
Make fake-mobile fail cleanly if font missing.

jlmjlm:


zenorogue:
Thanks! HyperRogue originally used the VeraBd font, it was updated to DejaVuSans-Bold at some point -- it is the same font but with international support -- but apparently fake-mobile was not updated, so the correct fix here was to change the font name. (Also TTF_OpenFont was potentially called in two places -- fixed this too.)

PR #381, opened 2024-07-10
Select the keyboard push-subdirection with Tab.

jlmjlm:
Previously, the keyboard push-subdirection was selected irregularly based on the exact angle of the map on the screen.

Change it to be consistently clockwise or counterclockwise, switching with the Tab key.

Also, have the keyboard push-subdirection help display on all pushable monsters.  Previously, it'd only show on Palace Guards and Skeletons.


zenorogue:
Thanks -- I like the old style, though. Thus, Shift+TAB returns to the old style.

PR #382, opened 2024-08-02
Add language-zh.cpp

1iuh:

PR #383, opened 2024-08-05
add measure words and  fix missing translations in language-zh.cpp

1iuh:
Thanks to @Quuxplusone  review in #382, we've made some fixes based on the review.

### DIFF
--- a/langen.cpp
+++ b/langen.cpp
@@ -294,14 +294,14 @@ void langZH() {
   current_language = "ZH";
   static std::pair<const char *, const char *> ds[] = {
     #define S(a,b) { a, b },
-    #define N(a,b,c)
+    #define N(a,b,c,d)
     #include "language-zh.cpp"
     #undef N
     #undef S
     };
   static std::pair<const char *, noun2> ns[] = {
     #define S(a,b)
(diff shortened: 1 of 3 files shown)


Quuxplusone:
@1iuh, why did you close this? (To keep it from being merged before it's ready? 🙂)  It looks plausible to me, FWIW — looks like exactly what I was thinking of!

zenorogue:
Thanks! I am also not sure how to interpret why is this closed.

Also one legal remark: since the paid versions of HyperRogue (and potentially other RogueViz games) have components which are not open source (such as the Steam interface), I need your permission to include your changes there (otherwise it violates the GPL). What are your thoughts on this? (Posting a permission here is OK.)

Quuxplusone:
As long as we're talking legal: #382 included a big binary diff to `NotoSans-Regular.ttf`. I assume that was at least _expected_ (because I assume the former font file didn't include any Chinese glyphs), but I would have liked to see some discussion about what its deal was, before you merged it. (A) are the additions legally OK? (B) are the additions non-malicious, and how can we tell? (C) does that diff cause any visible changes to Latin text?

PR #384, opened 2024-08-06
add measure words and  fix missing translations in language-zh.cpp

1iuh:
#383 had some bugs, so I close it. It's all fixed now.

#### About the permission
We agree to use this changes in any way.
#### About NotoSans-Regular.ttf
I replaced the original NotoSans-Regular.ttf because it didn't contain Chinese characters.
I got it from https://github.com/notofonts/noto-cjk.
It is open source and widely used. So I think it's safe.
It uses the SIL Open Font License, which we've added in 2750c02 , so I think it's legit.

### DIFF
--- a/langen.cpp
+++ b/langen.cpp
@@ -294,14 +294,14 @@ void langZH() {
   current_language = "ZH";
   static std::pair<const char *, const char *> ds[] = {
     #define S(a,b) { a, b },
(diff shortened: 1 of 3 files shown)


zenorogue:
Looks great, I just have been busy with other things. Thanks!

PR #394, opened 2024-11-18
starts_with compilation failure see comments for potential patch

envolution:
Makefile forces std c++14, and this code wasn't compatible

### DIFF
--- a/dialogs.cpp
+++ b/dialogs.cpp
@@ -436,12 +436,12 @@ EX namespace dialog {
       if(i < isize(str)) {
         char *ch = &str[i];
         if(*ch == ')') continue;
-        if(starts_with(ch, "）")) continue;
+        if (strncmp(ch, "）", strlen("）")) == 0) continue;
         }
       if(i > 0) {
         char *ch = &str[i - utfsize_before(str,i)];
         if(*ch == '(') continue;
-        if(starts_with(ch, "（")) continue;
+        if (strncmp(ch, "（", strlen("（")) == 0) continue;
         }
       if(textwidth(siz, str.substr(last, i-last)) > xs) {
         if(lastspace == last) ls = last_i, last = last_i;


envolution:
```
--- a/hyperrogue/dialogs.cpp	2024-11-18 09:20:13.764151597 -0500
+++ b/hyperrogue/dialogs.cpp_patched	2024-11-18 09:25:39.920842852 -0500
@@ -531,6 +531,11 @@
     return a;
     }

+  EX bool starts_with(const char* str, const char* prefix) {
+      size_t prefix_len = strlen(prefix);
+      return strncmp(str, prefix, prefix_len) == 0;
+  }
+
   EX void draw_list_slider(int x, int yt) {
     int a = uishape();
```

maybe something like this

envolution:
```
In file included from hyper.cpp:99:
dialogs.cpp: In function ‘int hr::dialog::displayLong(std::string, int, int, bool)’:
dialogs.cpp:439:12: error: ‘starts_with’ was not declared in this scope
  439 |         if(starts_with(ch, "）")) continue;
      |            ^~~~~~~~~~~
dialogs.cpp:444:12: error: ‘starts_with’ was not declared in this scope
  444 |         if(starts_with(ch, "（")) continue;
      |            ^~~~~~~~~~~
```
if there's a helper function for starts_with, it's not in the git repo

Quuxplusone:
Oh I see, and the calls to `starts_with` were added only today in 0137175d4. Maybe @zenorogue just forgot to `git add` the changes to some header file. The function must already exist in his checkout (and it's not in `std` so he must have done exactly what you and I said above). :)

Quuxplusone:
> no i believe it's in std c++20, which is probably a good reason to avoid all of my suggestions

Nope, C++20 adds _member_ `s.starts_with(t)` and `sv.starts_with(t)`, but those only work on `string` and `string_view`. There's no `std::starts_with(charptr, othercharptr)` free function, and even if there were, Hyperrogue has no `using std::starts_with` to pull it into scope.  Commit 0137175d46c8caf3e751fc103dfbfafe6b8b6b22 simply doesn't build.

envolution:
I see, thanks for that.  c++ always does my head in, even more so when sleep deprived

zenorogue:
Yeah, I added such a function to `util.cpp` and forgot to add this change to the commit. (I did not even know that a method with the same name and functionality is in C++20.) Thanks for notifying, this should be fixed now.

PR #405, opened 2025-06-02
Fix #99

### LINKED ISSUE #99: Out-of-bounds array access
While trying to package HyperRogue as Flatpak app, the game crashes because of a failed assertion. Flatpak by default enables the `_GLIBCXX_ASSERTIONS` flag for compilation, which in turn inserts the runtime range checks/assertions into the binary. The code where the problem arises is this:

https://github.com/zenorogue/hyperrogue/blob/bcc96ffa1195dd058c046cc856056845ff5b4f48/drawing.cpp#L1960-L1963

When starting the game, at one point the `ptds` vector **has 538 entries and is accessed at position 538 (out-of-bounds)**. Right then `qp` and `qp0` look like this:

```
(gdb) print qp0
$3 = {0 <repeats 29 times>, 393 <repeats 17 times>, 521 <repeats 15 times>, 525 <repeats 17 times>, 
  527, 533, 533, 533, 534, 535, 535, 535, 535, 535, 536, 537 <repeats 16 times>, 538, 538, 538, 538, 
  538, 538, 538, 538, 538, 538}
(gdb) print qp
$10 = {0 <repeats 28 times>, 393 <repeats 17 times>, 521 <repeats 15 times>, 525 <repeats 17 times>, 
  527, 533, 533, 533, 534, 535, 535, 535, 535, 535, 536, 537 <repeats 16 times>, 
  538 <repeats 11 times>}
```

The exact assertion error message is this:
```
/usr/include/c++/9.2.0/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = std::unique_ptr<hr::drawqueueitem>; _Alloc = std::allocator<std::unique_ptr<hr::drawqueueitem> >; std::vector<_Tp, _Alloc>::reference = std::unique_ptr<hr::drawqueueitem>&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
```

(Sry to have this really low-level technical bug, but I didn't get through the logic of the code in a reasonable amount of time)

NamorNiradnug:
See https://github.com/zenorogue/hyperrogue/issues/99#issuecomment-2928542115.

### DIFF
--- a/drawing.cpp
+++ b/drawing.cpp
@@ -2630,7 +2630,7 @@ EX void drawqueue() {
 
   for(PPR p: {PPR::TRANSPARENT_WALL}) {
     int pp = int(p);
-    sort(ptds.begin() + qp0[int(p)], ptds.begin() + qp[int(p)], 
+    sort(ptds.begin() + qp0[pp], ptds.begin() + qp[pp], 
       [] (const unique_ptr<drawqueueitem>& p1, const unique_ptr<drawqueueitem>& p2) {
         return p1->subprio > p2->subprio;
         });
--- a/drawing.cpp
+++ b/drawing.cpp
@@ -2616,12 +2616,11 @@ EX void drawqueue() {
   if(GDIM == 2)
   for(PPR p: all_side_prios) {
     int pp = int(p);
-    if(qp0[pp] == qp[pp]) continue;
     for(int i=qp0[pp]; i<qp[pp]; i++) {
       auto& ap = (dqi_poly&) *ptds[i];
       ap.cache = xintval(ap.V * ap.intester);
       }
-    sort(&ptds[qp0[pp]], &ptds[qp[pp]], 
+    sort(ptds.begin() + qp0[pp], ptds.begin() + qp[pp], 
       [] (const unique_ptr<drawqueueitem>& p1, const unique_ptr<drawqueueitem>& p2) {
         auto& ap1 = (dqi_poly&) *p1;
         auto& ap2 = (dqi_poly&) *p2;
@@ -2631,8 +2630,7 @@ EX void drawqueue() {
 
   for(PPR p: {PPR::TRANSPARENT_WALL}) {
     int pp = int(p);
-    if(qp0[pp] == qp[pp]) continue;
-    sort(&ptds[qp0[int(p)]], &ptds[qp[int(p)]], 
+    sort(ptds.begin() + qp0[int(p)], ptds.begin() + qp[int(p)], 
       [] (const unique_ptr<drawqueueitem>& p1, const unique_ptr<drawqueueitem>& p2) {
         return p1->subprio > p2->subprio;
         });


Quuxplusone:
FWIW, I agree, this looks like a good fix.
Alternatively, one could replace `&ptds[...]` with `&ptds.data()[...]`.
The important thing is not to invoke `ptds`'s own `vector::operator[]` with the one-past-the-end index: we must use either iterator math (as in this patch) or pointer math (as in my alternative above) to get at the one-past-the-end address.

Quuxplusone:
> > `&ptds.data()[...]`.
> 
> This one would still cause UB because `&ptds.data()[offset]` is equivalent to `&*(data() + offset)`, i.e. dereferencing an invalid pointer (again, it is UB even without reading). Simply `data() + offset` does work, but IMHO using iterators is more civil.

Off the top of my head I was unsure `&p[n]` is any more UB than `p+n`, when `p` is a raw pointer. On the one hand, [expr.unary/1](https://eel.is/c++draft/expr.unary#op-1.sentence-4) seems clearly to say it is UB. OTOH, `-Warray-bounds` refuses to diagnose it at compile time, and `-fsanitize=undefined` refuses to complain about it at runtime: https://godbolt.org/z/azdnKTn8G But that isn't positive evidence that it _is_ defined; that observation can just as easily be explained by the hypothesis that the behavior is undefined, but pragmatically speaking _can't be diagnosed_ because the whole world relies on it.

The critical thing in this case is that we stop calling `vector::operator[](n)`, because _that's_ the thing that's _very clearly_ specified as UB, and which libc++ is _actually in practice_ complaining about at runtime now.

Anyway, yeah, I vote to ship this patch using iterators, as you wrote it. :)

NamorNiradnug:
> > > `&ptds.data()[...]`.
> > 
> > 
> > This one would still cause UB because `&ptds.data()[offset]` is equivalent to `&*(data() + offset)`, i.e. dereferencing an invalid pointer (again, it is UB even without reading). Simply `data() + offset` does work, but IMHO using iterators is more civil.
> 
> Off the top of my head I was unsure `&p[n]` is any more UB than `p+n`, when `p` is a raw pointer. On the one hand, [expr.unary/1](https://eel.is/c++draft/expr.unary#op-1.sentence-4) seems clearly to say it is UB. OTOH, `-Warray-bounds` refuses to diagnose it at compile time, and `-fsanitize=undefined` refuses to complain about it at runtime: https://godbolt.org/z/azdnKTn8G and it's even fine in constexpr evaluation: https://godbolt.org/z/q6jP5zej7 But none of that is positive evidence that it _is_ defined; that observation can just as easily be explained by the hypothesis that pragmatically speaking this UB _can't be diagnosed_ because the whole world relies on it.

I did some googling and found out that in *C99* `&*e` is equivalent to `e` and `&x[y]` is equivalent to `x + y` by definition and thus is not UB even for `NULL` or past-to-end pointers. I guess compilers assume this is not UB in C++ too in order to make the behavior consistent.

zenorogue:
Thanks! I have noticed it breaking on my computer, so I have implemented a similar fix.

PR #423, opened 2025-08-18
Fix #145: add a shading cheat

### LINKED ISSUE #145: Feature request: Haunted Woods shading cheat
Can a new cheat option be added for the Haunted Woods, similar to the Camelot cheat used during the tour? It would shade alternating tiles based on their distance to the guiding line, maybe similar to how the Ocean and Ivory Tower always look, and also maybe indicate the guiding line itself in a special manner. This would make it easier to wrap your head around how the land works, since it's the only equidistant with an "invisible" guiding line right now, and it would also be a nice substitute for spamming compasses down everywhere to figure out where you are.

josephcsible:


josephcsible:
> Looks good, although from reading the diff, it seems that it does not turn off when you exit the cheat mode, so it could be used to cheat outside the cheat mode. (Could be fixed by turning `shadingcheat` off if not `cheater`, similar to how it is done for `camelotcheat`, or by changing the checks into `shadingcheat && cheater`, possibly made into a function.)

Okay, now I turn `shadingcheat` back off the same place that `timerghost` and `gen_wandering` get turned back on.

> Also one legal remark (it seems I have not asked you about this in 2021 for some reason): since the paid versions of HyperRogue (and potentially other RogueViz games) have components which are not open source (such as the Steam interface), I need your permission to include your changes there (otherwise it violates the GPL). What are your thoughts on this? (Posting a permission here is OK.)

That's fine with me. Do you have a full list of such libraries? If so, I'll post the FSF's recommended wording from https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs with them filled in.

PR #487, opened 2025-11-20
Slow down Orb of Winter effect when flashing effects are off

josephcsible:


josephcsible:
The check failures were spurious mirror issues, unrelated to this change.

PR #521, opened 2026-09-09
Add Circuit Land -- AND/NAND/OR logic puzzles guarded by hyperbugs

metaweta:
## Summary

Adds a new land, **Circuit Land**, containing procedurally-generated AND/NAND/OR logic-gate puzzles. Reachable only from the Hive, with the hyperbugs as native monsters; unlocks once combined Royal Jelly + Black Lotus + Fulgurite ≥ 30.

Each circuit is grown like the mutant ivy in the Clearing: hex cells only, branching where the geometry has room, running wires where it doesn't, always producing exactly `2^n` inputs for `n` gate-levels on every root-to-leaf path. Player feeds a TRUE signal into an input by dropping a Dead Orb; when the root gate evaluates to TRUE, the circuit dissolves in a lightning arc and drops a Circuit Seed.

## Content

- `laCircuit` land, `itCircuitSeed` treasure, walls `waCircuitSeed / waCircuitAND / waCircuitNAND / waCircuitOR / waCircuitWire / waCircuitInput`.
- Depth scales with `items[itCircuitSeed]/3 + 2`, capped at 6 (63 gates / 64 inputs).
- Dead Orbs spawn at low density inside the land so puzzles are solvable even in tactical mode.
- Trivially-satisfied trees (e.g. lone NAND) are rejected at plant time and re-rolled.
- Lightning arcs persist over the dissolved tree from solve until the player's next move, mirroring the arc behavior in Land of Storms.

## Implementation

- `circuit.cpp` — new file. `place_and_grow` is a DFS with backtracking that decides gate-vs-wire per cell from candidate count; `build_tree` sets up the seed and grows the single top child. Uses `landflags` (not `mondir`) for parent-direction storage so scout hyperbugs walking over passable input cells can't sever the tree.
- `celldrawer.cpp` — flat-on-ground gate rendering with IEEE-style filled/outlined logic-gate symbols (proportions from `Digital_logic_gates.svg`, output side rotated toward the parent); dark-green wire cells with bright-green 120° bent conductors that clip to the shared-edge midpoint of parent/child.
- `graph-item.cpp` — Circuit Seed gem rendered white with thick black outline to match the gate visual.
- `graph-wall.cpp` / `graph.cpp` — PCB-green floor color, no floor darkening for `laCircuit`.
- `landgen.cpp` — planting hook + Dead Orb spawn; `landlock.cpp` — `land_over` entry and `laHive → laCircuit` transition rule; `classes.cpp` — tactical roster; `items.cpp` — `circuit::check` on drop/collect.

## Test plan

- [ ] Enter Circuit Land from Hive (unlocked after 30 combined jelly + lotus + fulgurite).
- [ ] Grab Dead Orbs from the map, drop on input tiles, verify circuit dissolves + drops seed when root evaluates TRUE.
- [ ] Confirm lightning arcs are visible on the dissolved tree until the next player move.
- [ ] Solve enough seeds for a depth-4+ circuit; confirm wires bridge tight geometry and every path still passes through the same number of gates.
- [ ] Tactical mode: land appears in the tactics roster and can be played without prior inventory.


zenorogue:
Thanks for the pull request! I think it feels rather out of place currently -- it uses hyperbolic geometry (as in, the whole binary tree fits easily) but the interaction with other HyperRogue features seems rather weak. Satisfying tree circuits is rather easy (just go from the root, if you have to satisfy AND just satisfy both inputs, to satisfy OR find the input that is easier to satisfy, etc.). So I am currently not accepting the pull request, and not sure how much this is possible to fix.

Anyway, some ideas/thoughts:
- I think the wires should be also displayed under the gates.
- I think the wires should display whether they are currently carrying the 0 signal or the 1 signal. (For this, every circuit tile should probably remember its current state, so that it is not recomputed for every tile every frame -- but that would also make the evaluation faster since only the path towards the root would have to be evaluated.)
- I think the dead orbs should not be dropped randomly, but rather, they should already be on some inputs, thus setting them to 1. So sometimes you could satisfy a circuit by removing dead orbs. (Of course still checking that the circuits are not satisfied on generation.)
- Since the gates are static images, it would be better to not use curvepoint/queuecurve, but program their shapes in polygons.cpp and use queuepoly. Also I think it would look better without such a large linewidth (the segments currently do not join correctly). NAND gates are currently smaller, I think they should be all of this size.
- It feels it could be interesting if the "wire" tiles were passable -- there is often a path going through the circuit only using wire/free tiles.
- Maybe some things could be added so that players can create interesting maps using circuits in the map editor (similar to how people create redstone constructions in Minecraft). Maybe the inputs should act as pressure plates, so they could be activated by player or monsters creating on them? Maybe there could be outputs that do something else  (e.g. close a door)? For this, probably wparam should be used for circuit element orientation rather than landparam (I see you are currently using wparam for something else but maybe this is no longer needed once the circuit is created).
- Tree circuits are rather easy to satisfy. As a comment: allowing the same wire to be an input to multiple gates would make it harder, but still in P due to the nature of hyperbolic geometry.

metaweta:
> I am currently not accepting the pull request, and not sure how much this is possible to fix.

Fair enough.  Thanks for looking.

